PR #4, opened 2024-01-15
Error Handling

3rob3:


simondsmason:
Verified: upstream ImmichFrame (ghcr.io/immichframe/immichframe:latest) works correctly with album-based filtering on Home Office photo frame via immichframe-all-album-test container.

Issue #7, opened 2024-02-16
Store API-Key instead of credentials


JW-CH:
Storing the username and password does not really sound good to me. Why are you not using a API-Key from immich? API-Keys can easily be created and removed and do not contain sensitive data.

3rob3:
I will look into this, the API documentation doesn't have a lot of info about using the API Key. Do you know if it is the same as "accessToken"? 
Funny enough I had this same thought this morning before I got out of bed, and woke up to this message :)

JW-CH:
Hey, I just saw your reply. I thought I'd do this on my own if it is OK.

3rob3:
Awesome! Great idea with the settings.xml.....you have no idea how many times I pushed mine with my credentials intact.....ugh. As you can probably tell I am new at Git.

JW-CH:
> Awesome! Great idea with the settings.xml.....you have no idea how many times I pushed mine with my credentials intact.....ugh. As you can probably tell I am new at Git.

Ugh in that case you really need to watch out! People can see the Password you have used! Have a look at this: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository

I am active on the immich discord (pinged you) maybe we could discuss that there :)

3rob3:
I merged your changes and created new builds. Thanks again for your help! I will probably post to reddit, what is your reddit username so I can credit you?

JW-CH:
Thank you. I am not really using Reddit so it is totally fine, no need to mention :)

PR #8, opened 2024-02-16
Move auth from username/password to api key

JW-CH:
Hey @3rob3, thank you for your repo. Excuse me for being cheeky and creating a PR to your repo. I like the idea of running my own photoframe on a Raspberry Pi and found your repo.

My changes:
I excluded the Settings.xml in the gitignore file. So no one will ever push their credentials by mistake. I also created a Settings.example.xml file instead. Maybe a new to-do in the README is to copy that file and rename it to Settings.xml.

My other changes were moving from having your precious username and password in the Settings.xml. Also, if a user logs into immich with OIDC (third party login), they do not have a password (I think). So instead they could just create an API-Key and use this instead. Feel free to check my changes and change them. If you disagree, that's OK as well :)

EDIT: I also changed the README :)


JW-CH:
Related to #7

PR #10, opened 2024-02-21
Ability to choose multiple albums as datasource

### LINKED ISSUE #9: Possible to set a specific album
It would be very usefull to set a specific album instead of the whole library.

So i could share an album with my wife and set it as source in the ImmichFrame. With this, my wife and myself are able to define the pictures of the frame.

JW-CH:
- Code and solution cleanup (move classes into separate files, renamings, etc.)
- Add ability to add album IDs to Settings.xml to use albums as datasource

Fixes #9

@3rob3 I had to add the Settings.xml to the solution again, not sure if you meant to fully remove it. I had problems with my debug folder not having a Settings.xml. Hope this is OK, feel free to make more changes.

### DIFF
--- a/ImmichFrame.Desktop/app.manifest
+++ b/ImmichFrame.Desktop/app.manifest
@@ -3,7 +3,7 @@
   <!-- This manifest is used on Windows only.
        Don't remove it as it might cause problems with window transparency and embeded controls.
        For more details visit https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests -->
-  <assemblyIdentity version="1.0.0.3" name="AvaloniaTest.Desktop"/>
+  <assemblyIdentity version="1.0.0.4" name="ImmichFrame.Desktop"/>
 
   <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
     <application>
--- a/ImmichFrame/App.axaml.cs
+++ b/ImmichFrame/App.axaml.cs
@@ -2,8 +2,6 @@
 using Avalonia.Controls.ApplicationLifetimes;
 using Avalonia.Data.Core.Plugins;
 using Avalonia.Markup.Xaml;
-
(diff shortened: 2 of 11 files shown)


JW-CH:
> Thanks again for your work, changes and refactoring look good!
> 
> * I'm at work so can't test (plus I don't use albums), how is the performance?
> * Do you see any issue with GetAlbumAssetIds returning a video asset? Or does "format=JPEG" in the ImageUrl handle that scenario?

1. Performance should not be an issue at all. The album's assets should only be loaded once per day (right now)
2. Good catch. Did you support video assets already? The format=JPEG might not work right now

3rob3:
> 2. Good catch. Did you support video assets already? The format=JPEG might not work right now
I (probably wrongly) assumed getRandom was only returning photos. If not I have a feeling the "format=JPEG" is what is handling this and just showing an image grab of the video. I can test this after work.

My other thought was to remove getRandom entirely and just call getAllAssets and process similarly to how you did in GetAlbumAssetIds. There have been a few requests for only showing favorites, this would be an easy place to append ?isFavorite=true.

JW-CH:
Also this PR is only for the ability to have albums as a datasource. Let's create another issue/pr for the favourite thing.

I'll double-check if video assets work and if not, I'll try to fix it :)

JW-CH:
Video assets are just displayed as an image. I'd create another issue for playing video assets. My code should work fine.

3rob3:
that's what I figured, thanks for verifying. I had no plans on adding video support

JW-CH:
@3rob3 
> Is there no way to get the album uuid from an album name? I am not seeing anything in the API documentation. That is surprising.

I was thinking about this too. You could call "getAllAlbums" and filter locally for the name. But of course you could get multiple albums with the same name so using the UID was the obvious choice to start with.

PR #11, opened 2024-03-01
Ability to store images locally and set a keep duration, code refactor

JW-CH:


JW-CH:
Not sure if you want me to update the version number when I create a PR

3rob3:
> Not sure if you want me to update the version number when I create a PR

I will have a look this weekend, busy day at work. Have you noticed any issues with image rotation after loading from disk (I wouldn't think so)? That was an issue with my original implementation directly accessing the Immich library folder.

JW-CH:
> > Not sure if you want me to update the version number when I create a PR
> 
> I will have a look this weekend, busy day at work. Have you noticed any issues with image rotation after loading from disk (I wouldn't think so)? That was an issue with my original implementation directly accessing the Immich library folder.

The trick is to release the stream properly.

PR #13, opened 2024-03-12
Ability to display memories

### LINKED ISSUE #12: Is it possible to support memories?
It would be nice to cycle through memories as shown on the web. I have looked at the API could you possibly use the "getMemoryLane" function? I don't actually know C#.

What a great project!

JW-CH:
Closes #12


JW-CH:
Forgot to handle no assets, I'll fix that with another commit :)

PR #18, opened 2024-03-13
OpenApi, Imagedesc

JW-CH:
Use OpenApi to generate client, display ExifInfo Desc if no ImageDesc was set


JW-CH:
Still WIP, did not check everything.

PR #23, opened 2024-03-15
Change calls to generated API via OpenApi

JW-CH:


JW-CH:
@3rob3 could you check this too :) Build is working fine for me :)

Issue #28, opened 2024-03-17
Simple settings file validation


3rob3:
Simple manipulation/validation of setting values before loading, notify user if wrong. i.e.:
- validate numeric values for applicable settings
- all bool values present and bool
- Trim blanks and special character (trailing '/' in URL)
- no empty required values (URL, Api key)
- etc.

PR #34, opened 2024-03-25
Refactoring and Settings error handling/validation

JW-CH:


ImmichFrame/Helpers/AssetHelper.cs
3rob3, line 1: Everything else looks great, but I am a little confused my GetFilteredAssetIds. It seems like the intent is to combine using Memories/Albums/People (because they are ifs and not else ifs, and because you renamed OnlyMemories to ShowMemories)? Or are you just preparing for this possibility in the future?

JW-CH: I was thinking about combining them. OnlyMemories was renamed to ShowMemories because I think it's odd to configure albums and then not showing them. For me, it seems more user-friendly if you combine all the options together, exactly how the config was set up.

JW-CH:
Closes #28

JW-CH:
Still todo:
- Trim empty spaces
- API-Key settings validation?
- People/Albums settings validation
- Other settings validation

PR #35, opened 2024-03-27
Use Immich-Logo as icon

JW-CH:


3rob3:
I swear we/I did this already? I wonder what happened.

PR #45, opened 2024-03-29
configurable transition duration. Closes Feature requests #42

### LINKED ISSUE #42: Feature requests
1) Setting to set a time in seconds, how long a transition from one picture to another
2) Setting to be able to configure to show 2 vertical pictures at the same time next to each other

3rob3:
Closes #42
I could not get data binding to work for the CrossFade Duration so did it in Onloaded. 
Removed "if(_appsettings == null)" because it will be caught now in Settings.cs parse/validate.

### DIFF
--- a/ImmichFrame/Views/MainView.axaml.cs
+++ b/ImmichFrame/Views/MainView.axaml.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Threading.Tasks;
+using Avalonia.Animation;
 using Avalonia.Controls;
 using Avalonia.Interactivity;
 using Avalonia.Media.Imaging;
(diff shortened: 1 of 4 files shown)


ImmichFrame/Views/MainView.axaml.cs
```diff
@@ -43,24 +44,21 @@ private async void OnLoaded(object? sender, RoutedEventArgs e)
             ShowSplash();
             _appSettings = Settings.CurrentSettings;
             _viewModel.Settings = _appSettings;
+            //couldn't get data binding on CrossFade Duration to work, had to do this
```
JW-CH, line 47: Any more Info about this? Seems strange that this would not work.

3rob3: Nope, I fought with this for hours. Tried using different datatypes, tried converters, etc. with no luck. Seems like a limitation or bug in Avalonia. You can try, it won't even build if you try to set a binding on that property.

JW-CH: Did you raise the NotifyPropertyChanged event when setting the value?

3rob3: Sorry, I have been sick with covid since Friday afternoon. The property is TransitionDuration, it is a new member of the settings class just like all the other ones. The issue is not that it won't update the value, it just won't build at all when you set a binding on the CrossFade Duration.

JW-CH: Get well soon. I'll do some testing.

PR #48, opened 2024-04-06
Android support

3rob3:


BlakeFL:
Android! LETS GO!

JW-CH:
image
image
@BlakeFL some progress on the android version :)

BlakeFL:
Looks great.  I need to start making a case for an old Kindle Fire tablet to prepare for this.

ImmichFrame/Models/Settings.cs
```diff
@@ -38,13 +43,88 @@ public static Settings CurrentSettings
...
+            {
+                using (XmlWriter writer = XmlWriter.Create(sww))
+                {
+                    xsSubmit.Serialize(writer, this);
+                    xml = sww.ToString();
+                }
+            }
+
+            // TODO: some error handling with try catch?
+            File.WriteAllText(AppDomain.CurrentDomain.BaseDirectory + @"Settings.xml", xml);
```
3rob3, line 127: I'd like to re-think this serialize section in the future. I'd prefer to keep XML and the built-in app settings separate.

JW-CH: This is still separate, or am I missing something?

3rob3: Overwriting the XML.

3rob3: My preference would be if the XML is there then use it, don't modify it, and potentially don't even allow settings screen to be shown. If XML is not there, then use app settings and allow settings screen, same as Android.

hunkyn:
I have an old fire tablet and the apk is failing. Is the a requirement for OS version or something that should be met?

3rob3:
> I have an old fire tablet and the apk is failing. Is the a requirement for OS version or something that should be met?
minSdkVersion="28", which is Android 9. What version is yours?

hunkyn:
> > I have an old fire tablet and the apk is failing. Is the a requirement for OS version or something that should be met?
> 
> minSdkVersion="28", which is Android 9. What version is yours?
> 
> 

This is fire tablet and it has the latest version. Is there a way i can see the SDK version for fireOS?

JW-CH:
> > > I have an old fire tablet and the apk is failing. Is the a requirement for OS version or something that should be met?
> > 
> > 
> > minSdkVersion="28", which is Android 9. What version is yours?
> 
> This is fire tablet and it has the latest version. Is there a way i can see the SDK version for fireOS?
> 
> EDIT: Seems like this fire tablet does not support Android 9. Atleast fireOS.

We could try to lower the required OS version. We'll check if lower android versions could be supported.

PR #54, opened 2024-04-15
Handle assets without a thumbnail / thumb hash

### LINKED ISSUE #53: Error: Value cannot be Null (Parameter 's')
Hi, I've got multiple times the error 

```
Value cannot be Null (Parameter 's')

```

After pressing ok it's working again... but my frame has no input 

image

JW-CH:
Some images (mostly live photos/generated videos) do not have thumbnails. ImmichFrame displays thumbnails. This handles this and skips the asset.

Fixes #53

### DIFF
--- a/ImmichFrame/Helpers/AssetHelper.cs
+++ b/ImmichFrame/Helpers/AssetHelper.cs
@@ -59,8 +59,8 @@ namespace ImmichFrame.Helpers
             }
 
             if (assetsAdded)
-                // return only unique assets, no duplicates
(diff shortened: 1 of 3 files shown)


3rob3:
I haven't fully committed to Immich yet, and all my stuff still goes to Google Photos first, then comes down (manually) and is imported into Immich monthly. I wonder if this is why I have never seen this issue?

JW-CH:
image
I had mostly snapchat assets that were not able to generate a thumbnail. Also the video part of motion photos do not get thumbnails according to bo0tzz on discord.

PR #57, opened 2024-04-17
Refactor MainView

JW-CH:
MainView Cleanup -> Logic in ViewModel, Dispose old view when navigating

### DIFF
--- a/ImmichFrame/Views/MainView.axaml.cs
+++ b/ImmichFrame/Views/MainView.axaml.cs
@@ -50,12 +50,13 @@ public partial class MainView : BaseView
             if (_appSettings == null)
                 throw new SettingsNotValidException("Settings could not be parsed.");
 
-            await _viewModel.ShowNextImage();
 
             if (transitioningControl.PageTransition is CrossFade crossFade)
             {
                 crossFade.Duration = TimeSpan.FromSeconds(_appSettings.TransitionDuration);
             }
+
+            await _viewModel.ShowNextImage();
 
             _viewModel.TimerEnabled = true;
             timerImageSwitcher = new System.Threading.Timer(_viewModel.NextImageTick, null, 0, _appSettings.Interval * 1000);
--- a/ImmichFrame/Views/MainView.axaml.cs
+++ b/ImmichFrame/Views/MainView.axaml.cs
@@ -8,8 +8,6 @@ using ImmichFrame.Helpers;
 using ImmichFrame.Models;
 using ImmichFrame.ViewModels;
 using System;
-using System.Linq;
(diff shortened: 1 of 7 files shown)


ImmichFrame/Views/MainView.axaml.cs
```diff
@@ -44,31 +38,34 @@ private async void OnLoaded(object? sender, RoutedEventArgs e)
...
+            timerImageSwitcher = new System.Threading.Timer(_viewModel.NextImageTick, null, 0, _appSettings.Interval * 1000);
             if (_appSettings.ShowClock)
             {
-                timerLiveTime = new System.Threading.Timer(timerLiveTimeTick, null, 0, 1000); //every second
+                timerLiveTime = new System.Threading.Timer(_viewModel.LiveTimeTick, null, 0, 1 * 1000); //every second
             }
             if (_appSettings.ShowWeather)
             {
-                timerWeather = new System.Threading.Timer(timerWeatherTick, null, 0, 600000); //every 10 minutes
+                timerWeather = new System.Threading.Timer(_viewModel.WeatherTick, null, 0, 10 * 60 * 1000); //every 10 minutes
```
3rob3, line 68: unrelated to your change, but i think we should update the weather timer to 5 minutes (or even less). I only used 10 minutes because everyone used the same API key, and I didn't want to hit the limit. Now everyone has own API key.

JW-CH: Do you think that weather change info is needed every 5 minutes? I was thinking that 10 is too much

JW-CH:
Somehow it feels buggy/not that responsive. I'll have a detailed look into it again.

ImmichFrame/Views/MainView.axaml.cs
3rob3: Personally....yeah. The weather around here changes pretty frequently. I have noticed it lagging behind a few times.

3rob3:
> Somehow it feels buggy/not that responsive. I'll have a detailed look into it again.

After this PR? I can't try it until I get home.

JW-CH:
> > Somehow it feels buggy/not that responsive. I'll have a detailed look into it again.
> 
> After this PR? I can't try it until I get home.

Yea, it feels choppy. It might be the `timerImageSwitcher?.Change(_appSettings!.Interval * 1000, _appSettings!.Interval * 1000);` but I am not sure.

3rob3:
> Somehow it feels buggy/not that responsive. I'll have a detailed look into it again.

I tried this branch on Windows and did not notice any difference, seemed normal. Were you on Android emulator (I didn't try there)?

PR #60, opened 2024-04-25
fix license in readme

abe-101:


3rob3:
Nice, thanks! I didn't even think about this.

PR #62, opened 2024-04-25
Interactive Control, Settings Quit button

### LINKED ISSUE #58: Feature: Interactive control
I know it is meant to be used as a picture frame.

But would it be possible to e.g. if you use it at a tablet (Linux or maybe Android in future) to show your pictures to others that they can interact like following?
- freeze slide show to current frame to look more in detail at it
- go back to last picture
- go further to next picture

Might be touch or cursor keys on keyboard maybe.

### LINKED ISSUE #61: Feature: Add Exit button to settings page
If valid settings are not entered, there is no way to exit the application.

3rob3:


3rob3:
Closes #58, Closes #61

abe-101 (approved):
Thanks!

ImmichFrame/ViewModels/MainViewModel.cs
```diff
@@ -176,7 +180,17 @@ public async Task ShowPreviousImage()
             await ShowMessageBoxFromThread(ex.Message, "");
         }
     }
-
+    public async void PauseImageAction()
+    {
+        //ResetTimer?.Invoke(this, new EventArgs());
+        await PauseImage();
+    }
+    public Task PauseImage()
```
JW-CH, line 188: Why does this method return a task?

ImmichFrame/Views/SettingsView.axaml
```diff
@@ -25,18 +25,23 @@
...
 			<vm:SettingsViewModel />
 		</Design.DataContext>
-		<Grid DockPanel.Dock="Bottom" ColumnDefinitions="*,*">
+		<Grid DockPanel.Dock="Bottom" ColumnDefinitions="*,*,*">
 			<Button Grid.Column="0"
-					Content="Cancel"
-					HorizontalAlignment="Stretch"
-					Command="{Binding CancelCommand}"
-					HorizontalContentAlignment="Center"/>
+              Content="Cancel"
```
JW-CH, line 30: Fix formatting? :)

ImmichFrame/Views/SettingsView.axaml
```diff
@@ -25,18 +25,23 @@
...
+              HorizontalAlignment="Stretch"
+              Command="{Binding CancelCommand}"
+              HorizontalContentAlignment="Center"/>
 			<Button Grid.Column="1"
-					Content="Save"
-					HorizontalAlignment="Stretch"
-					Command="{Binding SaveCommand}"
-					HorizontalContentAlignment="Center"/>
-		</Grid>
+              Content="Save"
```
JW-CH, line 35: Also Formatting

JW-CH (changes requested):

ImmichFrame/ViewModels/MainViewModel.cs
3rob3: Because it doesn't contain any awaits the method can't be async. This allows it to be awaited. How would you do it?

ImmichFrame/Views/SettingsView.axaml
3rob3: Are you looking at it in VS? Looks correct to me.

ImmichFrame/Views/SettingsView.axaml
3rob3: same

ImmichFrame/ViewModels/MainViewModel.cs
JW-CH: Why does it need to be awaited? Just return void and remove the await in the call.

ImmichFrame/Views/SettingsView.axaml
JW-CH: Nope, was looking at it online.
image

3rob3: edited my comment, think its fixed now.

PR #75, opened 2024-05-13
Margin setting

### LINKED ISSUE #74: Feature: Adjustable margins
Add settings value for Margin. This can allow fine tuning for display alignment or adjust for overscan

3rob3:
Added Margin setting
Added NumericUpDown margins to align with other controls (unclear why this needs explicitly set, stackpanel should have done this)
Closes #74

### DIFF
--- a/ImmichFrame/Settings.example.xml
+++ b/ImmichFrame/Settings.example.xml
@@ -1,69 +1,72 @@
 <Settings>
-	<ImmichServerUrl>http://YourIpOrUrl:2283</ImmichServerUrl>
-	<!--Your immich API-Key, see here https://immich.app/docs/features/command-line-interface#obtain-the-api-key-->
-	<ApiKey>YourApiKey</ApiKey>
+  <ImmichServerUrl>http://YourIpOrUrl:2283</ImmichServerUrl>
+  <!--Your immich API-Key, see here https://immich.app/docs/features/command-line-interface#obtain-the-api-key-->
+  <ApiKey>YourApiKey</ApiKey>
 
-	<!--OPTIONAL: Choose one or multiple Albums to have in the rotation-->
-	<!--Remove this section if you want to show random images-->
-	<Albums>
-		<Album>First Album UID</Album>
-		<Album>Second Album UID</Album>
-	</Albums>
+  <!--Adjustable margins (left,top,right,bottom) for fine tuning position-->
+  <Margin>0,0,0,0</Margin>
 
-	<!--OPTIONAL: Choose one or multiple People to have in the rotation-->
-	<!--Remove this section if you want to show random images-->
-	<People>
-		<Person>First Person UID</Person>
-		<Person>Second Person UID</Person>
-	</People>
+  <!--OPTIONAL: Choose one or multiple Albums to have in the rotation-->
+  <!--Remove this section if you want to show random images-->
+  <Albums>
+    <Album>First Album UID</Album>
+    <Album>Second Album UID</Album>
+  </Albums>
 
-	<!--Interval for photo cycling in seconds-->
-	<Interval>8</Interval>
+  <!--OPTIONAL: Choose one or multiple People to have in the rotation-->
+  <!--Remove this section if you want to show random images-->
+  <People>
+    <Person>First Person UID</Person>
+    <Person>Second Person UID</Person>
+  </People>
 
-	<!--Photo fade in/out time in seconds-->
-	<TransitionDuration>1.0</TransitionDuration>
+  <!--Interval for photo cycling in seconds-->
+  <Interval>8</Interval>
 
-	<!--Determines if images should be saved locally-->
-	<DownloadImages>false</DownloadImages>
+  <!--Photo fade in/out time in seconds-->
+  <TransitionDuration>1.0</TransitionDuration>
 
-	<!--Show todays memories-->
-	<ShowMemories>false</ShowMemories>
+  <!--Determines if images should be saved locally-->
+  <DownloadImages>false</DownloadImages>
 
-	<!--Duration for images to keep if existing, renews the image if older than the duration. Duration in Days-->
-	<RenewImagesDuration>20</RenewImagesDuration>
+  <!--Show todays memories-->
+  <ShowMemories>false</ShowMemories>
 
-	<!--Show clock in lower left bottom-->
-	<ShowClock>true</ShowClock>
-	<ClockFontSize>48</ClockFontSize>
-	<!--The format of the clock (ex. h:mm tt). See here for options https://www.c-sharpcorner.com/blogs/date-and-time-format-in-c-sharp-programming1-->
-	<ClockFormat>h:mm tt</ClockFormat>
+  <!--Duration for images to keep if existing, renews the image if older than the duration. Duration in Days-->
+  <RenewImagesDuration>20</RenewImagesDuration>
 
-	<!--Show the date photo was taken.-->
-	<ShowPhotoDate>true</ShowPhotoDate>
-	<PhotoDateFontSize>36</PhotoDateFontSize>
-	<!--The format of the date (ex. MM/dd/yyyy). See here for options https://www.c-sharpcorner.com/blogs/date-and-time-format-in-c-sharp-programming1-->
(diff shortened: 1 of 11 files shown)


ImmichFrame/Settings.example.xml
```diff
@@ -1,69 +1,72 @@
 <Settings>
-	<ImmichServerUrl>http://YourIpOrUrl:2283</ImmichServerUrl>
```
JW-CH, line 2: Once again, the settings format was messed up.

ImmichFrame/Views/MainView.axaml
```diff
@@ -20,7 +20,7 @@
   <Design.DataContext>
     <vm:MainViewModel />
   </Design.DataContext>
-  <Grid ColumnDefinitions="*, *, *" RowDefinitions="*, *, *">
+  <Grid ColumnDefinitions="*, *, *" RowDefinitions="*, *, *" Margin="{Binding Settings.Margin}">
```
JW-CH, line 23: Don't add this to all the views, we have the `WindowView` that could be that one place to set this. (The wrapper view). With that, all the views will have the needed `margin`.

ImmichFrame/Views/SettingsView.axaml
```diff
@@ -85,11 +85,11 @@
         </StackPanel>
         <StackPanel>
           <TextBlock Text="Interval"/>
-          <NumericUpDown Value="{Binding Settings.Interval}" FormatString="F0"/>
+          <NumericUpDown Value="{Binding Settings.Interval}" FormatString="F0" Margin="10,0" />
```
JW-CH, line 88: Can you do the margin here as a style? Not really related tho this PR.

ImmichFrame/ViewModels/SettingsViewModel.cs
```diff
@@ -16,6 +17,8 @@ public partial class SettingsViewModel : NavigatableViewModelBase
 
         [ObservableProperty]
         private ObservableCollection<ListItem> albumList;
+        [ObservableProperty]
+        private string margin;
```
JW-CH, line 21: Why adding a separate property? Just use `Settings.Margin`

ImmichFrame/Views/SettingsView.axaml
```diff
@@ -101,15 +101,19 @@
         </StackPanel>
         <StackPanel>
           <TextBlock Text="RenewImagesDuration"/>
-          <NumericUpDown Value="{Binding Settings.RenewImagesDuration}" FormatString="F0"/>
+          <NumericUpDown Value="{Binding Settings.RenewImagesDuration}" FormatString="F0" Margin="10,0" />
+        </StackPanel>
+        <StackPanel>
+          <TextBlock Text="Margin"/>
+          <TextBox Text="{Binding Margin}" />
```
JW-CH, line 108: Use `Settings.Margin`.

ImmichFrame/Views/SettingsView.axaml
JW-CH: Also Add the `FormatString` to the style

ImmichFrame/ViewModels/SettingsViewModel.cs
```diff
@@ -48,6 +51,7 @@ public SettingsViewModel()
 
             PeopleList = new ObservableCollection<ListItem>(Settings.People.Select(x => new ListItem(x.ToString())));
             AlbumList = new ObservableCollection<ListItem>(Settings.Albums.Select(x => new ListItem(x.ToString())));
+            Margin = new string(Settings.Margin.ToString());
```
JW-CH, line 54: The `.ToString()` already creates a string. I'd never create a new string with `new string()`. Also, this should not be needed since you can access `Settings.Margin`.

ImmichFrame/ViewModels/SettingsViewModel.cs
```diff
@@ -93,6 +97,7 @@ public void SaveAction()
             {
                 Settings.People = PeopleList.Select(x => Guid.Parse(x.Value)).ToList();
                 Settings.Albums = AlbumList.Select(x => Guid.Parse(x.Value)).ToList();
+                Settings.Margin = Thickness.Parse(Margin);
```
JW-CH, line 100: Use `Settings.Margin`.

JW-CH (changes requested):

JW-CH:
Ah and btw: The `Margin` on the `NumericUpDown` is needed, because the `Textblock` has a style.

ImmichFrame/Settings.example.xml
3rob3: Must be a region thing, it is not messed up:
image

ImmichFrame/Views/SettingsView.axaml
3rob3: Agreed, good idea.

ImmichFrame/Settings.example.xml
JW-CH: Your formatting adds two spaces, mine adds a tabstop

JW-CH: image
what are your c# settings here? :)

ImmichFrame/ViewModels/SettingsViewModel.cs
3rob3: Settings.Margin is a Thickness, Margin is a string.

ImmichFrame/ViewModels/SettingsViewModel.cs
3rob3: Forgot to remove that, was troubleshooting

ImmichFrame/ViewModels/SettingsViewModel.cs
3rob3: Settings.Margin is a Thickness. You can't bind a TextBox text to a Thickness without a converter, this was far less code.

ImmichFrame/Views/SettingsView.axaml
3rob3: Settings.Margin is a Thickness.

ImmichFrame/Settings.example.xml
3rob3: Whatever the defaults are, in this case same as your screenshot.

3rob3:
> Ah and btw: The `Margin` on the `NumericUpDown` is needed, because the `Textblock` has a style.

Yeah, dumb mistake. Fixed.

ImmichFrame/Views/MainView.axaml
3rob3: Good catch, didn't think about that. Problem is I'm not sure how to do it :). It doesn't work setting it in the UserControl tag. It works if you set a Style for Grid, but SettingsView parent container is a DockPanel, not a grid.
How were you thinking?

JW-CH: image
This works for me

JW-CH: Maybe we also could set the background to black?

3rob3: Yeah, that works for me too, but not when you set the binding. It just does nothing.

3rob3: Try my branch with that change if you can, dev.3rob3.MarginControl

JW-CH:
I made a change to your PR, can you review that too?

3rob3:
> I made a change to your PR, can you review that too?

It works, but only updates Margin once (in the MainWindowViewModel constructor). So now you have to quit and re-open to get the new margin to apply.

JW-CH:
> > I made a change to your PR, can you review that too?
> 
> It works, but only updates Margin once (in the MainWindowViewModel constructor). So now you have to quit and re-open to get the new margin to apply. It also doesn't catch errors for invalid Margin values entered, it will just crash out and make you enter all your settings again. Moving the Margin stuff from the constructor to the Navigate method solves the first issue.
> 
> Also, I added the language to date/time change (its just a few lines in MainViewModel). This is for the issue bo0tzz submitted.

I created a new branch for the proper error page thing. I noticed that too. Other stuff will break too if you provide invalid settings.

3rob3:
So should I merge this then? What did you think about moving Margin= into Navigate?

PR #78, opened 2024-05-15
feat: better error page to display exceptions, update to FluentUI

JW-CH:


ImmichFrame/Views/SettingsView.axaml
```diff
@@ -3,180 +3,181 @@
...
+		<Style Selector="TextBox">
+			<Setter Property="Margin" Value="10,3,10,10"/>
+			<Setter Property="VerticalAlignment" Value="Center"/>
+		</Style>
+		<Style Selector="CheckBox">
+			<Setter Property="Margin" Value="10,0,10,10"/>
+			<Setter Property="VerticalAlignment" Value="Center"/>
+		</Style>
+		<Style Selector="NumericUpDown">
+			<Setter Property="Margin" Value="10,3,10,10"/>
```
3rob3, line 27: I'm not sure why, but the NumericUpDown's are not properly aligned anymore. This seems to fix it:
<Setter Property="Margin" Value="0,3"/>

ImmichFrame/Views/ErrorView.axaml
```diff
@@ -0,0 +1,46 @@
...
+							Command="{Binding SettingsCommand}"/>
+						<Button
+							Content="Continue"
+							Command="{Binding ContinueCommand}"/>
+					</StackPanel>
+				</StackPanel>
+			</Border>
+		</Grid>
+	</DockPanel>
+</UserControl>
```
3rob3, line 46: Love it on Desktop. Unfortunately it is not working on Android (ErrorView isn't shown at all).

JW-CH: The ErrorView was working, just not showing on startup (When the settings were not correct). If you for example have a wrong ApiKey, the page will show. I made it more responsive, see my Image:
image

ImmichFrame/Views/SettingsView.axaml
JW-CH: I'm still not happy with the settings page. I will rework it in another PR.

ImmichFrame/Views/ErrorView.axaml
3rob3: Ah, ok. I should have tried creating a different type of error.

PR #82, opened 2024-05-27
Create an album for recent ImmichFrame displayed photos

### LINKED ISSUE #80: Create an album on immich server "Recent ImmichFrame Photos"
### Description

via bo0tzz on Discord:
I just had an interesting idea, I wonder what y'all think of it. When you have a frame going somewhere, a picture might pass by that you want to do something with, and depending on the setup the frame might not have a way to interact with it. How about having an option to put the latest n pictures into an album automatically, so you can easily pull them up from another device?

### Use Case

Easily access/Interact/do something with a photo you recently saw displayed on your phot frame.

### Details

_No response_

### Pre-Submission Checklist

3rob3:
Closes #80

### DIFF
--- a/ImmichFrame/App.axaml
+++ b/ImmichFrame/App.axaml
@@ -1,18 +1,18 @@
 <Application xmlns="https://github.com/avaloniaui"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              x:Class="ImmichFrame.App"
-             RequestedThemeVariant="Default"
-			 xmlns:local="using:ImmichFrame.Views"
-			 xmlns:sty="using:FluentAvalonia.Styling"
-			 xmlns:ui="using:FluentAvalonia.UI.Controls"
-			 xmlns:uip="using:FluentAvalonia.UI.Controls.Primitives">
-             <!-- "Default" ThemeVariant follows system theme variant. "Dark" or "Light" are other available options. -->
-	
-    <Application.Styles>
-		<sty:FluentAvaloniaTheme />
-		<StyleInclude Source="avares://ImmichFrame/Views/Templates/SettingsSection.axaml" />
-    </Application.Styles>
-	<Application.DataTemplates>
-		<local:ViewLocator/>
-	</Application.DataTemplates>
+             RequestedThemeVariant="Dark"
+             xmlns:local="using:ImmichFrame.Views"
+             xmlns:sty="using:FluentAvalonia.Styling"
+             xmlns:ui="using:FluentAvalonia.UI.Controls"
+             xmlns:uip="using:FluentAvalonia.UI.Controls.Primitives">
+  <!-- "Default" ThemeVariant follows system theme variant. "Dark" or "Light" are other available options. -->
+
+  <Application.Styles>
+    <sty:FluentAvaloniaTheme />
+    <StyleInclude Source="avares://ImmichFrame/Views/Templates/SettingsSection.axaml" />
+  </Application.Styles>
+  <Application.DataTemplates>
+    <local:ViewLocator/>
+  </Application.DataTemplates>
 </Application>
--- a/ImmichFrame/App.config
+++ b/ImmichFrame/App.config
@@ -79,11 +79,8 @@
             <setting name="Margin" serializeAs="String">
                 <value>0,0,0,0</value>
             </setting>
-            <setting name="UseImmichFrameAlbum" serializeAs="String">
-                <value>False</value>
-            </setting>
             <setting name="ImmichFrameAlbumName" serializeAs="String">
-                <value>ImmichFrame(DeviceName)</value>
+                <value />
             </setting>
         </ImmichFrame.Properties.Settings>
     </userSettings>
--- a/ImmichFrame/App.config
+++ b/ImmichFrame/App.config
@@ -79,6 +79,12 @@
             <setting name="Margin" serializeAs="String">
                 <value>0,0,0,0</value>
             </setting>
+            <setting name="UseImmichFrameAlbum" serializeAs="String">
+                <value>False</value>
+            </setting>
+            <setting name="ImmichFrameAlbumName" serializeAs="String">
+                <value>ImmichFrame(DeviceName)</value>
+            </setting>
         </ImmichFrame.Properties.Settings>
     </userSettings>
 </configuration>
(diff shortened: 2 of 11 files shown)


ImmichFrame/App.config
```diff
@@ -79,6 +79,12 @@
             <setting name="Margin" serializeAs="String">
                 <value>0,0,0,0</value>
             </setting>
+            <setting name="UseImmichFrameAlbum" serializeAs="String">
```
JW-CH, line 82: As with the API-Key settings some time ago. Wouldn't one name setting be enough? Like if you set a name, that implies that you want to use the ImmichFrame Album?

ImmichFrame/Helpers/AssetHelper.cs
```diff
@@ -35,6 +37,55 @@ public class AssetHelper
...
             return await GetRandomAsset();
         }
+        public async Task AddAssetToAlbum(AssetResponseDto assetToAdd)
+        {
+            using (var client = new HttpClient())
+            {
+                var settings = Settings.CurrentSettings;
+                client.UseApiKey(settings.ApiKey);
+                var immichApi = new ImmichApi(settings.ImmichServerUrl, client);
+                var immichAlbums = await immichApi.GetAllAlbumsAsync(null, null);
```
JW-CH, line 47: Loading all albums when adding a photo seems odd to me. Could you store the album ID somehow, or at least cache it? (Cache it for all the requests)

ImmichFrame/Views/MainView.axaml.cs
```diff
@@ -55,7 +55,8 @@ private async void OnLoaded(object? sender, RoutedEventArgs e)
             {
                 crossFade.Duration = TimeSpan.FromSeconds(_appSettings.TransitionDuration);
             }
-
+            AssetHelper _assetHelper = new AssetHelper();
+            await _assetHelper.DeleteImmichFrameAlbum();
```
JW-CH, line 59: It might be better to delete the "old" album on startup. If you close the app, you might still want to be able to see the last images in that album. Also, if the app crashes, the album will not be deleted.

A nice feature would be to have the last X images in there, if the app for example runs for a few days, you get huge albums without this feature. Not sure if this is possible.

JW-CH (changes requested):

ImmichFrame/App.config
3rob3: Agreed, I will make this change. 
FYI:
Original implementation was just a checkbox, and album name was auto-created (ImmichFrame{machine name)) but I couldn't get it working on Android.

ImmichFrame/Views/MainView.axaml.cs
3rob3: This is on startup, not sure what you mean here.

For the second point, if I understand you correctly I do this already. I am only storing a maximum of 50 assets in the album, once you hit that I remove the oldest from the album, then add new one.

JW-CH:
I did not manage to test these changes right now. Will do when I get time. I am unsure about the album loading in the AssetHelper

ImmichFrame/Models/Settings.cs
```diff
@@ -268,10 +271,6 @@ private static Settings ParseSettings(Dictionary<string, object> SettingsValues)
                 case "ShowPhotoDate":
                 case "ShowImageDesc":
                 case "ShowImageLocation":
-                    if (!bool.TryParse(value.ToString(), out var boolValue))
```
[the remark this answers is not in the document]
3rob3: OMG....oops.

ImmichFrame/Helpers/AssetHelper.cs
```diff
@@ -35,6 +37,56 @@ public class AssetHelper
...
+        }
+        public async Task DeleteImmichFrameAlbum()
+        {            
+            using (var client = new HttpClient())
+            {
+                var settings = Settings.CurrentSettings;
+                client.UseApiKey(settings.ApiKey);
+                var immichApi = new ImmichApi(settings.ImmichServerUrl, client);
+                var immichAlbums = await immichApi.GetAllAlbumsAsync(null, null);
+                var immichFrameAlbum = immichAlbums.FirstOrDefault(album => album.AlbumName == settings.ImmichFrameAlbumName);
```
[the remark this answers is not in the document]
3rob3: Only deletes Album if there is one with same name as specified in ImmichFrameAlbumName. This is needed so we don't orphan images in that album. Only other option is to GetAllAbums, GetAllAssets in the album, remove asset for the album ( one at a time). End result is the same, just a lot more resources spent doing it.

3rob3:
> I did not manage to test these changes right now. Will do when I get time. I am unsure about the album loading in the AssetHelper

I have been thinking about how to improve this. I feel like it is more risk to remove it. Weird stuff could happen if someone deletes the album manually. It works well with negligible performance hit.

ImmichFrame/Models/Settings.cs
```diff
@@ -299,6 +298,9 @@ private static Settings ParseSettings(Dictionary<string, object> SettingsValues)
                 case "Language":
                     property.SetValue(settings, value);
                     break;
+                case "ImmichFrameAlbumName":
```
JW-CH, line 305: Can you consolidate this with Language and Clockformat? If the cases do the same, have them together.

3rob3: done. I did that on purpose so in the future we could validate language.

ImmichFrame/App.axaml
```diff
@@ -1,18 +1,18 @@
...
-             <!-- "Default" ThemeVariant follows system theme variant. "Dark" or "Light" are other available options. -->
-	
-    <Application.Styles>
-		<sty:FluentAvaloniaTheme />
-		<StyleInclude Source="avares://ImmichFrame/Views/Templates/SettingsSection.axaml" />
-    </Application.Styles>
-	<Application.DataTemplates>
-		<local:ViewLocator/>
-	</Application.DataTemplates>
+             RequestedThemeVariant="Dark"
```
JW-CH, line 4: Good catch, did not know this was an option.

ImmichFrame/Helpers/AssetHelper.cs
3rob3: Updated, can you approve (if you approve).

JW-CH (approved):

PR #86, opened 2024-06-11
API Changes version 1.106.1

JW-CH:
Use the new API introduced in v1.106.1

### DIFF
--- a/ImmichFrame/Helpers/AssetHelper.cs
+++ b/ImmichFrame/Helpers/AssetHelper.cs
@@ -137,7 +137,7 @@ namespace ImmichFrame.Helpers
                 foreach (var lane in memoryLane)
                 {
                     var assets = lane.Assets.ToList();
-                    assets.ForEach(asset => asset.ImageDesc = lane.Title);
+                    assets.ForEach(asset => asset.ImageDesc = $"{lane.YearsAgo} years ago");
 
                     allAssets.AddRange(assets);
                 }
--- a/ImmichFrame/Helpers/AssetHelper.cs
+++ b/ImmichFrame/Helpers/AssetHelper.cs
@@ -137,7 +137,7 @@ namespace ImmichFrame.Helpers
                 foreach (var lane in memoryLane)
                 {
                     var assets = lane.Assets.ToList();
-                    assets.ForEach(asset => asset.ImageDesc = $"{lane.YearsAgo} years ago");
+                    assets.ForEach(asset => asset.ImageDesc = $"{lane.YearsAgo} {(lane.YearsAgo == 1 ? "year" : "years")} ago");
 
                     allAssets.AddRange(assets);
                 }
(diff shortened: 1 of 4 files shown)


ImmichFrame/Helpers/AssetHelper.cs
```diff
@@ -137,7 +137,7 @@ private async Task<IEnumerable<AssetResponseDto>> GetMemoryAssets()
                 foreach (var lane in memoryLane)
                 {
                     var assets = lane.Assets.ToList();
-                    assets.ForEach(asset => asset.ImageDesc = lane.Title);
+                    assets.ForEach(asset => asset.ImageDesc = $"{lane.YearsAgo} years ago");
```
jrasm91, line 140: not sure if you have to handle 1 year special case = "1 year ago" vs X = "X years ago".

jrasm91 (approved):

ImmichFrame/Helpers/AssetHelper.cs
3rob3: Good catch. Easy solution is:
 assets.ForEach(asset => asset.ImageDesc = $"{lane.YearsAgo} year(s) ago");
 What do you think of that?

JW-CH: tbh I was just lazy, will handle it properly

ImmichFrame/Models/Settings.cs
```diff
@@ -127,6 +156,19 @@ private void Validate()
...
     }
 
+    private void LoadAssetFormats()
+    {
+        using (var client = new HttpClient())
+        {
+            var api = new ImmichApi(this.ImmichServerUrl, client);
+            client.UseApiKey(ApiKey);
+
+            var cfg = api.GetConfigAsync().Result;
```
JW-CH, line 166: The `GetConfig`-Endpoint requires admin permission. This will not work then... trying to figure out a good way to get information about the format

jrasm91: Why do you need to know the formats in the first place?

ImmichFrame/Models/AssetResponseDto.cs
```diff
@@ -87,14 +88,23 @@ private async Task<Stream> DownloadImageAsync(string localPath)
...
+                contentType = data.Headers["Content-Type"].ToString();
+            }
 
             var stream = data.Stream;
             var ms = new MemoryStream();
             stream.CopyTo(ms);
             ms.Position = 0;
             if (Settings.CurrentSettings.DownloadImages)
             {
+                var ext = contentType?.ToLower() == "image/webp" ? "webp" : ".jpeg";
```
jrasm91, line 105:
```suggestion
                var ext = contentType?.ToLower() == "image/webp" ? "webp" : "jpeg";
```

JW-CH: Thanks

3rob3:
This looks good from what I can review on web. I won't be able to test until later today, plus I'm moving my immich server to new hardware so that will take some time. I will leave it up to you if you want to push a release sooner, i'm good with it if you have tested.

PR #93, opened 2024-07-09
Convert to Generated Regex, remove \d usage

crobibero:
Generated Regex: https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators

From https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#decimal-digit-character-d

> \d matches any decimal digit. It is equivalent to the \p{Nd} regular expression pattern, which includes the standard decimal digits 0-9 as well as the decimal digits of a number of other character sets.


3rob3:
Looks good, thanks for contributing!

PR #95, opened 2024-07-10
Add remote control navigation to Settings screen Closes #91

### LINKED ISSUE #91: AndroidTV Remote doesn't work in settings
### Version

v1.0.5.0

### Deployment Method

Chromecast with Google TV

### Description

Opening the app on first install opens a settings view that is impossible to navigate without mouse/keyboard.

I was able to work around this by using an ADB mouse and keyboard app on my phone, but if I enter settings after configuration I am stuck in the settings view as back doesn't close the view.

### Reproduction

1. Install app on AndroidTV
2. Attempt to use remote to navigate

### Expectations

The user should be able to navigate settings using the dpad on the remote

### Configuration

_No response_

### Logs

_No response_

### Pre-Submission Checklist

- [X] This is a bug report and not a feature request

- [X] I have provided all of the required information to reproduce the bug (config, logs, etc.)

3rob3:
Closes #91


JW-CH:
Could you also add a Settings.examlpe.json? Also make sure that the Settings.json does not get pushed!

JW-CH:
I did not manage to test it in time, sorry. But it looks solid after skipping over the code. I'll be gone for the weekend so I won't be able to fully review.

3rob3:
> Could you also add a Settings.examlpe.json? Also make sure that the Settings.json does not get pushed (Gitignore)!

Settings.json is just a blank template. It is set to Content so it gets copied to the output folder. If the user doesn't fill it out we will error and allow them to enter via Settings screen, or they can quit and do it manually. If it doesn't exist we will create it. I'm not sure I see a reason for Settings.example anymore.

3rob3:
> I did not manage to test it in time, sorry. But it looks solid after skipping over the code. I'll be gone for the weekend so I won't be able to fully review.

Thanks for reviewing. I have been testing it pretty thoroughly.

crobibero:
This looks like the user would be forced to fully reconfigure the app after update?

crobibero:
I installed this on my AndroidTV for testing and can navigate on the error view, but hitting the select button doesn't do anything

3rob3:
> I installed this on my AndroidTV for testing and can navigate on the error view, but hitting the select button doesn't do anything

Did you build it yourself? This is still in a PR and hasn't been released yet. Should be sometime this week.

crobibero:
Yes, I built main with this branch merged in to test the functionality

3rob3:
> Yes, I built main with this branch merged in to test the functionality

ughhh. I'm seeing same thing. I had this working.

JW-CH:
I like the new Settings.json, but we should provide some kind of "migration" from an XML-File to the new JSON. This should not be hard to do since we can just check for an XML-File, load it, save it as JSON-File and delete the old XML-File.

3rob3:
> I like the new Settings.json, but we should provide some kind of "migration" from an XML-File to the new JSON. This should not be hard to do since we can just check for an XML-File, load it, save it as JSON-File and delete the old XML-File.

That would be nice, just seems like it’s more effort than it’s worth since it’s just a one time thing. My opinion might be off since I have probably recreated that file 100 times now 😀. 

FYI I got things working with .NET 8 and updated Avalonia. I will push some more changes soon.

ImmichFrame.Android/ImmichFrame.Android.csproj
```diff
@@ -1,14 +1,14 @@
...
+    <TargetFramework>net8.0-android</TargetFramework>
     <Nullable>enable</Nullable>
     <AndroidPackageFormat>apk</AndroidPackageFormat>
-    <AndroidEnableProfiledAot>True</AndroidEnableProfiledAot>
+    <RunAOTCompilation>true</RunAOTCompilation>
   </PropertyGroup>  
 
   <ItemGroup>
-    <PackageReference Include="Avalonia.Android" Version="$(AvaloniaVersion)" />
+    <PackageReference Include="Avalonia.Android" Version="11.1.0-rc2" />
```
JW-CH, line 11: is this specific version needed?

3rob3: Yes, until the official 11.1 release is out. Hopefully soon.

3rob3: Official release is out, updated

PR #101, opened 2024-08-01
Reverting version of Microsoft.Extensions.ApiDescription.Client and NSwag.ApiDescription.Client packages.

ScottFries:
The update to this package from 1382fdbd8e91185a7f01fe6b7fd26b73eebf2cd8 was causing an issue with URLs not being assembled correctly.


ScottFries:
For any trying to replicate/test, note that changing package versions and/or cleaning solution is not enough to remove the cached version of the respective generated code files.

PR #108, opened 2024-08-05
Add current weather image next to temperature

3rob3:


ScottFries:
Very nice touch! Honestly with this I'm going to drop the textual report for weather state; might be a better logical default, but that could definitely just be me.

3rob3:
> Very nice touch! Honestly with this I'm going to drop the textual report for weather state; might be a better logical default, but that could definitely just be me.

dang it, will need another setting to be able to do that.

PR #113, opened 2024-08-07
Refactor MainView code to MainViewModel, make async where possible

3rob3:


ImmichFrame/ViewModels/MainViewModel.cs
```diff
@@ -15,30 +16,58 @@ namespace ImmichFrame.ViewModels;
...
+        NextImageCommand = new RelayCommand(async () => await NextImageAction());
+        PreviousImageCommand = new RelayCommand(async () => await PreviousImageAction());
         PauseImageCommand = new RelayCommand(PauseImageAction);
+        QuitCommand = new RelayCommand(ExitApp);
         NavigateSettingsPageCommand = new RelayCommand(NavigateSettingsPageAction);
     }
 
+    public async Task InitializeAsync()
+    {
+        ShowSplash();
```
JW-CH, line 50: await ShowSplash?

ImmichFrame/ViewModels/MainViewModel.cs
```diff
@@ -91,6 +120,12 @@ public async Task SetImage(AssetResponseDto asset, Stream? preloadedAsset = null
         }
 
     }
+    private void ShowSplash()
```
JW-CH, line 123: make async

ImmichFrame/ViewModels/MainViewModel.cs
```diff
@@ -91,6 +120,12 @@ public async Task SetImage(AssetResponseDto asset, Stream? preloadedAsset = null
         }
 
     }
+    private void ShowSplash()
+    {
+        var uri = new Uri("avares://ImmichFrame/Assets/Immich.png");
+        var bitmap = new Bitmap(AssetLoader.Open(uri));
+        SetImage(bitmap);
```
JW-CH, line 127: await SetImage()

JW-CH (changes requested):

ImmichFrame/ViewModels/MainViewModel.cs
3rob3: There is nothing asynchronous happening with ShowSplash call stack so I think it is unnecessary. Ultimately all that is happening is setting 'Image' directly to a bitmap. I also considered just removing it altogether.

PR #115, opened 2024-08-08
- Highlight required settings with red border (if blank). - Avalonia update

3rob3:


3rob3:
What do you guys think of this. If Immich URL or API key are blank they will be highlighted red. Looks like this:
image

ScottFries:
> What do you guys think of this. If Immich URL or API key are blank they will be highlighted red. Looks like this: image @ScottFries

Seems like a nice clarifying improvement to UX!

JW-CH:
I like the highlighting, but I don't like that nothing really happens. It is red but still lets you save corrupt/invalid settings.

3rob3:
> I like the highlighting, but I don't like that nothing really happens. It is red but still lets you save corrupt/invalid settings.

updated.

PR #116, opened 2024-08-11
Android: expose and call PauseImage when app loses focus but isn't closed. 

3rob3:


ImmichFrame.Android/MainActivity.cs
```diff
@@ -16,15 +17,35 @@ namespace ImmichFrame.Android;
...
         return base.CustomizeAppBuilder(builder)
             .WithInterFont();
     }
+    protected override void OnStart()
+    {
+        base.OnStart();
+        if (!isAppRunning)
+        {
+            isAppRunning = true;
+            MainViewModel.Instance?.PauseImage();
```
ScottFries, line 39: Bit awkward having `PauseImage` serve the roll of both `Pause` & `Resume`. Makes logic like this confusing to read and leaves no room for state validation since we just blindly toggle the current state.

3rob3: PauseImage isn't new code, I just exposed it. Anyway, isn't that exactly what a pause does? Not sure why it's awkward.

3rob3: Didn't work anyway, deleting.

PR #117, opened 2024-08-11
1. Don't create new task if already running and screensaver starts. 2. Add singleTop launchMode to AndroidManifest 3. Add uses-feature leanback to AndroidManifest
2. Add singleTop launchMode to AndroidManifest
3. Add uses-feature leanback to AndroidManifest

3rob3:


JW-CH (approved):
LGTM

PR #121, opened 2024-08-22
Directly use Settings.CurrentSettings

3rob3:
Don't create new instances of Settings everywhere, just directly use Setting.CurrentSettings (with the exception of ViewModels where the properties need to be observable).


JW-CH (approved):
👍🏻

PR #122, opened 2024-08-22
Nicer weather icons loaded from assets instead of http requests

3rob3:
OpenWeatherMap icons kinda sucked :).  Including them in project also speeds up load and reduces http requests. Also solves SSL issue on Smartview devices.


JW-CH (approved):
👍🏻

PR #143, opened 2024-09-24
Use search endpoint for random assets & cache optimizations

JW-CH:
> [!WARNING]
This will only work with immich-server version 1.116 or higher

Following things were changed:
- updated Openapi-Specs
- use new random endpoint
- use caching for random endpoint to reduce load on immich api
- use search metadata endpoint instead of deprecated `GetPersonAssets`
- updated generation of immich client

### DIFF
--- a/ImmichFrame.Core/Logic/ImmichFrameLogic.cs
+++ b/ImmichFrame.Core/Logic/ImmichFrameLogic.cs
@@ -18,6 +18,20 @@ namespace ImmichFrame.Core.Logic
         private DateTime lastFilteredAssetRefesh;
         private List<Guid> ImmichFrameAlbumAssets = new List<Guid>();
         private static AlbumResponseDto immichFrameAlbum = new AlbumResponseDto();
+
+
+        private Task<IEnumerable<Guid>> _excludedAlbumAssets;
+        private Task<IEnumerable<Guid>> ExcludedAlbumAssets
+        {
+            get
+            {
+                if(_excludedAlbumAssets == null)
+                    _excludedAlbumAssets = GetExcludedAlbumAssets();
+
+                return _excludedAlbumAssets;
+            }
+        }
+
         private Task<Dictionary<Guid, AssetResponseDto>?> FilteredAssetInfos
         {
             get
@@ -125,11 +139,11 @@ namespace ImmichFrame.Core.Logic
                 // Exclude videos
                 list = list.Where(x => x.Type != AssetTypeEnum.VIDEO);
 
-                var excludedalbumAssetIds = (await GetExcludedAlbumAssets()).Select(x => x.Id);
+                var excludedList = await ExcludedAlbumAssets;
 
                 // Exclude assets if configured
-                if (excludedalbumAssetIds.Any())
-                    list = list.Where(x => !excludedalbumAssetIds.Contains(x.Id));
+                if (excludedList.Any())
+                    list = list.Where(x => !excludedList.Contains(Guid.Parse(x.Id)));
 
                 // return only unique assets, no duplicates, only with Thumbnail
                 return list.Where(x => x.Thumbhash != null).DistinctBy(x => x.Id).ToDictionary(x => Guid.Parse(x.Id));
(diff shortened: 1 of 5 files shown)


JW-CH:
Do not merge before the release of immich v1.116

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -161,7 +161,7 @@ private async Task<IEnumerable<AssetResponseDto>> GetMemoryAssets()
                 foreach (var lane in memoryLane)
```
3rob3, line 161: This is in reference to lines 40 and 43 (but I can't comment there). Should we just disable nullable at the project level? We are not really using it correctly anyway (AFAIK).

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -227,9 +227,26 @@ private async Task<IEnumerable<AssetResponseDto>> GetPeopleAssets()
...
+                                Type = AssetTypeEnum.IMAGE
+                            };
+                            var personInfo = await immichApi.SearchMetadataAsync(metadataBody);
+
+                            total = personInfo.Assets.Total;
+
+                            allAssets.AddRange(personInfo.Assets.Items);
+                            page++;
+                        }
+                        while (total == batchSize);
```
3rob3, line 249: Personal preference, but I find it more readable to not use "do", and just put the while line in it's place.

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -227,9 +227,26 @@ private async Task<IEnumerable<AssetResponseDto>> GetPeopleAssets()
                 {
                     try
                     {
-                        var personInfo = await immichApi.GetPersonAssetsAsync(personId);
-
-                        allAssets.AddRange(personInfo);
+                        int page = 1;
```
3rob3, line 230: Can we just set these (page and batchSize) directly in metadataBody since we don't use them elsewhere?

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -254,37 +271,46 @@ private async Task<IEnumerable<AssetResponseDto>> GetPeopleAssets()
 
             return filteredAssetInfos.ElementAt(rnd).Value;
         }
+
+        List<AssetResponseDto> RandomAssetList = new List<AssetResponseDto>();
         private async Task<AssetResponseDto?> GetRandomAsset()
         {
+            if (RandomAssetList.Any())
+            {
+                // TODO: Check somehow for excluded albums?
```
3rob3, line 280: So Albums and ExcludedAlbums is unchanged for now, correct?

3rob3:
Nice job.

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
JW-CH: Will do :)

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
JW-CH: We have to look into this, but not in this PR?

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
JW-CH: I removed the excluded album check because we have to query for albums separately. And this would be silly if we cache the assets but query for every asset's albums. Probably a change on immich's api required.

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
JW-CH: Oh, I overlooked this. `page` is incremented and used to fetch everything. `batchSize` is used for the check if another batch should be loaded

3rob3: I guess I did to!

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
3rob3: yeah, was just a thought I had.

PR #153, opened 2024-10-07
Adaptive Android icons by @Blisterexe

3rob3:


Blisterexe:
this pr also zooms in the icon and makes it follow themed icon preferences

PR #167, opened 2024-10-16
Feat: Zoom in / out of faces

JW-CH:
Zoom in or out of (known) faces when available, also show faces in metadata.

This ignores faces that do not have a name (unassigned faces)


3rob3:
Neat!

PR #171, opened 2024-10-22
Feat: Display calendar

### LINKED ISSUE #170: Calendar overlay
### Description

Looking great so far!
Wondering if it would be possible to add in the ability to overlay a calendar from an ics file feed (URL). 

### Use Case

For larger displays this would allow users to display photos with the text calendar overlay. Alternatively, the calendar could go below to not be on top of the image, or on top with no background but the photo.

### Details

Options could include:
How many days to show
Background colors on individual event (light)
Event locations (at/in/none)
Event end time
Day x or Y label (for repeating events)
Font size
Current and future events only (hide items that are past)

### Pre-Submission Checklist

JW-CH:
closes #170


JW-CH:
image
not pretty at all but it works

PR #175, opened 2024-10-30
Feat: Shortcuts

JW-CH:
- Right arrow -> Next
- Left arrow -> Back (not implemented)
- Space -> Pause/Play


JW-CH:
@3rob3 FYI, implemented this :)

PR #178, opened 2024-11-01
Feat: Show memories in order

### LINKED ISSUE #163: Display Memories in correct order
### Description

Me and my family love the Memory-Lane of immich. So this project was the right solution to get the Pictures of the past years on our Home-Tablet as a daily slide-show. 
For me it would be great to have the option that the memory-slide-show was shown in the correct order (a year ago, two years ago... and so on).
At the moment (with "Albums" empty and only `"ShowMemories": true`) it shows all photos in a random order.

### Use Case

The Use Case would resemble the Memory-Lane Feature of Immich itself.

### Details

_No response_

### Pre-Submission Checklist

JW-CH:
Fixes #163 
Show memories in correct order if only memories are set up.


3rob3:
nice!

PR #180, opened 2024-11-04
Feat: Implement imageLocationFormat

3rob3:


immichFrame.Web/src/lib/components/elements/asset-info.svelte
```diff
@@ -13,13 +13,23 @@
...
+		$configStore.imageLocationFormat ?? 'City,State,Country',
+		asset.exifInfo?.city ?? '',
+		asset.exifInfo?.state ?? '',
+		asset.exifInfo?.country ?? ''
+	);
 
 	$: availablePeople = asset.people?.filter((x) => x.name);
+	function formatLocation(format: string, city?: string, state?: string, country?: string) {
+		const locationParts: string[] = [];
+		if (format.includes('City') && city) locationParts.push(city);
```
JW-CH, line 28: Thank you for the implementation. Could you split the format string by ',' and loop over the results? For example if someone has `city, city` as format, it just displays city once. Also if `nocity` is the format, city will be displayed. I am not saying that his makes perfect sense, but just implementation wise this makes more sense.

immichFrame.Web/src/lib/components/elements/asset-info.svelte
```diff
@@ -13,13 +13,23 @@
 
 	$: time = new Date(assetDate);
 
-	$: formattedDate = format(time, $configStore.photoDateFormat ?? 'dd.MM.yyyy');
+	$: formattedDate = format(time, $configStore.photoDateFormat ?? 'dd.MM.yyyy');
```
JW-CH, line 16: Are you using VSCode? If yes, maybe you want to install the `ESLint`-Extension so that this does not happen :) also I might add a lint check to the PR build checks.

JW-CH: By 'this' I mean the empty spaces ofc :)

JW-CH (changes requested):

immichFrame.Web/src/lib/components/elements/asset-info.svelte
3rob3: ESLint doesn't catch that. I forget to 'cntrl-k,cntrl-d' though.

immichFrame.Web/src/lib/components/elements/asset-info.svelte
3rob3: Can you check again?

PR #184, opened 2024-11-11
Feat: Reimplement Fontcolor-Setting

JW-CH:
- Use user defined primary color
- Fallback to default if none is defined

### DIFF
--- a/immichFrame.Web/src/lib/stores/config.store.ts
+++ b/immichFrame.Web/src/lib/stores/config.store.ts
@@ -5,6 +5,9 @@ function createConfigStore(settings: WebClientSettings) {
   const { subscribe, set } = writable(settings);
 
   function ps(settings: WebClientSettings) {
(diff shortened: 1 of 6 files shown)


immichFrame.Web/src/lib/stores/config.store.ts
```diff
@@ -5,6 +5,9 @@ function createConfigStore(settings: WebClientSettings) {
   const { subscribe, set } = writable(settings);
 
   function ps(settings: WebClientSettings) {
+    if (settings.fontColor) {
+      document.documentElement.style.setProperty('--primary-color', settings.fontColor);
```
3rob3, line 9: Is this the right place to set this? Seems like other stuff like this is set in home-page.svelte. I'm just asking, you know way more about this project than i do.

JW-CH: You are right-ish? Haha, I moved it. My thought was to set it where the settings are loaded, but I think the home-page is the place where this fits.

PR #188, opened 2024-11-16
docs: specify transition duration as a float

mmomjian:
Based on my testing, durations of 0.1 or 0.5 seconds seem to work great.


JW-CH:
Nice catch, thanks!

PR #190, opened 2024-11-16
Basic webhook implementation

JW-CH:


ImmichFrame.Core/Helpers/JsonConverters.cs
```diff
@@ -0,0 +1,38 @@
...
+        public override TInterface Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override void Write(Utf8JsonWriter writer, TInterface value, JsonSerializerOptions options)
+        {
+            if (value == null)
+            {
+                JsonSerializer.Serialize(writer, value, options);
```
3rob3, line 17: I am unclear why we are serializing this and then just returning without using it. Shouldn't we either just return, or do something with it?

JW-CH: This is just a converter example from the web. This will just return the serialized value (null). Not really my code, but I thought I'd leave it as is.

PR #192, opened 2024-11-17
Feat: Add layout setting

JW-CH:
- Add layout setting
- Cleanup documentation


README.md
```diff
@@ -62,71 +62,49 @@ ImmichFrame is easy to run on your desired plattform. Get the latest stable rele
...
+| [Metadata](#metadata)   | ShowPhotoDate              | boolean             | true                 | Displays the date of the current image.                                                                               |
+| [Metadata](#metadata)   | PhotoDateFormat            | string              | yyyy-MM-dd           | Date format.                                                                                                          |
+| UI                      | FontColor                  | string              | #FFFFFF              | Lets you choose a primary color for your UI.                                                                          |
+| UI                      | Splitview                  | single \| splitview | splitview            | Allow two portrait images to be displayed next to each other                                                          |
+| Misc                    | ImmichFrameAlbumName       | string              |                      | \*Client only. Creates album and stores last 100 photos displayed.                                                    |
```
3rob3, line 94: I was just in Readme yesterday and looking at this section, then got overwhelmed. Thanks. We should probably remove all Client stuff from here though. I think we should also remove all settings that aren't implemented in Web. I was afraid to that before talking to you.

docker/Settings.example.json
```diff
@@ -1,37 +1,38 @@
 {
```
3rob3, line 1: What do you think of dropping Settings.json? ENV is much easier to use (for me) and generally how other Dockers work.

3rob3: FYI, not saying for this PR, just a comment

PR #199, opened 2024-11-19
Feat: Clock & Metadata styling

### LINKED ISSUE #198: Request: Option to add semi-transparent background to overlayed text
### Description

It can be hard to ensure text like the clock and especially the metadata is legible against any photo. I suggest building in the option to have a semitransparent background behind whatever text elements are configured to increase legibility.

### Use Case

For when any overlayed text is configured that you want to ensure is always legible for a specific font color, regardless of the photo being displayed.

### Details

The color of the background could also be configurable, and the background could have a gradient fade into full transparency outside of the text regions.

### Pre-Submission Checklist

JW-CH:
> [!Warning]
> FontColor will be renamed to PrimaryColor

- rename FontColor to PrimaryColor
- add SecondaryColor setting
- add style setting: `none`, `solid`, `transition`, `blur`

Closes #198


docker/example.env
```diff
@@ -19,7 +19,9 @@ ShowImageDesc=true
 ShowPeopleDesc=true
 ShowImageLocation=true
 ImageLocationFormat=City,State,Country
-FontColor=#FF5733
+PrimaryColor=#FF5733
```
3rob3, line 22: Should we default these to the default soft white color? Is it #f5deb3? Red is very jarring

Install_Web.md
```diff
@@ -67,7 +67,7 @@ services:
       DownloadImages: "false"
       RefreshAlbumPeopleInterval: "12"
       # UI
-      FontColor: "#FF5733"
+      PrimaryColor: "#FF5733"
       # Misc
       ImmichFrameAlbumName: ""
```
3rob3, line 72: this isn't implemented, is it?

3rob3:
Looks good! Only minor comments, nothing stopping the merge

Install_Web.md
JW-CH: Nope, it is not

PR #205, opened 2024-11-26
Base font size

3rob3:
implement a configurable base font size.


Install_Web.md
```diff
@@ -71,6 +71,7 @@ services:
       RefreshAlbumPeopleInterval: "12"
       # UI
       PrimaryColor: "#FF5733"
+      BaseFontSize: "medium"
```
JW-CH, line 74: Why is this 'medium' and the README default says 17px?

docker/example.env
```diff
@@ -25,6 +25,7 @@ ImageLocationFormat=City,State,Country
 PrimaryColor=#FF5733
 SecondaryColor=#000000
 Style=none
+BaseFontSize=medium
```
JW-CH, line 28: Dito

README.md
```diff
@@ -100,6 +100,7 @@ ImmichFrame is easy to run on your desired plattform. Get the latest stable rele
 | UI                      | SecondaryColor             | string                              | #000000              | Lets you choose a secondary color for your UI. (Only used with `style=solid or transition`) Use hex with alpha value to edit opacity. |
 | UI                      | Style                      | none \| solid \| transition \| blur | none                 | Background-style of the clock and metadata.                                                                                           |
 | UI                      | Layout                     | single \| splitview                 | splitview            | Allow two portrait images to be displayed next to each other                                                                          |
+| UI                      | BaseFontSize               | string                              | 17px                 | Sets the base font size, use standard CSS formats                                                                                     |
```
JW-CH, line 103: Can you maybe add a link to some explaination? For example this: https://developer.mozilla.org/en-US/docs/Web/CSS/font-size

(standard CSS formats)[https://developer.mozilla.org/en-US/docs/Web/CSS/font-size]

Install_Web.md
JW-CH: Also I am not sure about the naming `BaseFontSize`, but I don't really have a better name.

JW-CH (changes requested):

JW-CH:
Btw: I really like the approach. Thank you :)

Install_Web.md
3rob3: I did that to show the different options. I was afraif epople would think you had to use numbers. Open to suggestions.....

docker/example.env
3rob3: any of these are allowed:
https://developer.mozilla.org/en-US/docs/Web/CSS/font-size

README.md
3rob3: I was going to add that exact link, was afraid it wouldn't fit.

Install_Web.md
3rob3: Your default was 17px, which is larger than "medium"

README.md
JW-CH: `(standard CSS formats)[https://developer.mozilla.org/en-US/docs/Web/CSS/font-size]`

this will make a link to mozilla with the text of 'standard CSS formats'. Could you change that :)

3rob3: ooh, i like it. Sure

Install_Web.md
JW-CH: > Also I am not sure about the naming `BaseFontSize`, but I don't really have a better name.

Should we think about a better name for this setting?

3rob3: seemed good to me, what don't you like about it? It's can't be "FontSize" because that's not accurate, I assume others are based off of it with a multiplier (i.e clock)?

JW-CH (approved):

PR #216, opened 2024-12-03
Feat: Server side image caching

JW-CH:
Adding server side caching again to the webapi. This will download new images into a `ImageCache`-Folder. Images will be refreshed after the number of days configured in the `RenewImagesDuration`-Setting. You can enable this features with the `DownloadImages`-Setting.


3rob3 (approved):
looks good to me

PR #218, opened 2024-12-03
Bug: Date format not using locale

JW-CH:
The date formatting was not using the language/locale defined in the config.


3rob3 (approved):
looks good to me!

PR #227, opened 2024-12-09
Add text-shadow to app.css to improve readability

iCanHazSkillz:
Addresses suggestion made in request #198

### DIFF
--- a/immichFrame.Web/src/app.css
+++ b/immichFrame.Web/src/app.css
@@ -59,6 +59,5 @@
 html {
 	height: 100%;
 	width: 100%;
-	font-size: 17px;
-	text-shadow: 2px 2px 4px rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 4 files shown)


JW-CH:
Thank you for your contribution. We are mostly using tailwind for our CSS. What are your thoughts on a custom CSS file that you can map to your container to edit/override and customize your immichFrame?

iCanHazSkillz:
Thank you! I went ahead and created 2 classes to accommodate shadows on both small and large text and applied it to the clock as well.

Let me know if there should be anything else added.

JW-CH:
I think there happened something when pulling from main. Your code contains some of the new stuff but not everything and does not build anymore :P

JW-CH:
Replaced by #232

3rob3:
this looks really nice, thanks! A lot easier to see/read now.

PR #228, opened 2024-12-10
Create new API for getting images outside of WebView.

3rob3:


ImmichFrame.WebApi/ImmichFrame.WebApi.csproj
```diff
@@ -5,7 +5,7 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <SpaRoot>..\immichFrame.Web</SpaRoot>
-    <SpaProxyLaunchCommand>npm run dev</SpaProxyLaunchCommand>
+    <SpaProxyLaunchCommand>npm run dev -- --host</SpaProxyLaunchCommand>
```
JW-CH, line 8: Is this line needed in the default?

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +39,43 @@ public async Task<FileResult> GetImage(Guid id)
...
+        [Produces("application/json")]
+        public async Task<IActionResult> GetRandomImage()
+        {
+            var _settings = new WebClientSettings();
+            var randomImage = await _logic.GetNextAsset() ?? throw new AssetNotFoundException("No asset was found");
+
+            //var randomImageUrl = Url.Action("GetRandomImage", null, new { id = randomImage.Id }, Request.Scheme);
+            var image = await _logic.GetImage(new Guid(randomImage.Id));
+            string randomImageBase64;
+            using (MemoryStream memoryStream = new MemoryStream())
```
JW-CH, line 53: Image already returns a stream, why is a memory stream needed here?

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +39,43 @@ public async Task<FileResult> GetImage(Guid id)
...
+            using (var memoryStream = (MemoryStream)randomImage.ThumbhashImage!)
+            {
+                byte[] byteArray = memoryStream.ToArray();
+                thumbHashBase64 = Convert.ToBase64String(byteArray);
+            }
+            CultureInfo cultureInfo = new CultureInfo(_settings.Language);
+            string photoDate = randomImage.LocalDateTime.ToString(_settings.PhotoDateFormat, cultureInfo) ?? string.Empty;
+            string imageDesc = randomImage.ImageDesc ?? string.Empty;
+
+            return Ok(new
```
JW-CH, line 72: I don't like the `Ok()`. Can't you just return a class/tuple?

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +39,43 @@ public async Task<FileResult> GetImage(Guid id)
...
+            {
+                await image.fileStream.CopyToAsync(memoryStream);
+
+                byte[] fileBytes = memoryStream.ToArray();
+
+                randomImageBase64 = Convert.ToBase64String(fileBytes);
+            }
+
+            string thumbHashBase64;
+            using (var memoryStream = (MemoryStream)randomImage.ThumbhashImage!)
```
JW-CH, line 63: Can't you just access `randomImage.Thumbhash` here?

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +39,43 @@ public async Task<FileResult> GetImage(Guid id)
 
             return File(image.fileStream, image.ContentType, image.fileName); // returns a FileStreamResult
         }
+
+        [HttpGet("RandomImage", Name = "GetRandomImage")]
+        [Produces("application/json")]
+        public async Task<IActionResult> GetRandomImage()
```
JW-CH, line 45: Return actual class/object instead of action result.

JW-CH: Also, maybe change the naming? This returns more than just an image (image, metadata)

ImmichFrame.WebApi/Controllers/AssetController.cs
JW-CH: Maybe return an imagestream and the full `AssetResponseDto`?

JW-CH (changes requested):

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: > I don't like the `Ok()`. Can't you just return a class/tuple?

this is how am ensuring we get an http Ok response. Seemed like good practice Why do you "not like" it?

ImmichFrame.WebApi/ImmichFrame.WebApi.csproj
3rob3: > Is this line needed in the default?

Impossible to debug from another device on network without it. whats the issue?

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: fileStream is not seekable, or directly convertible to byte array. This was only way I got it to work. Feel free to suggest an alternative.

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: > Can't you just access `randomImage.Thumbhash` here?

Yes I can fix

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: > Return actual class/object instead of action result.

this is how am ensuring we get an http Ok response. Seemed like good practice

3rob3: > Also, maybe change the naming? This returns more than just an image (image, metadata)

done

ImmichFrame.WebApi/ImmichFrame.WebApi.csproj
JW-CH: > > Is this line needed in the default?
> 
> Impossible to debug from another device on network without it. whats the issue?

You are right. This is debug only. I just thought we keep it minimal and if you need to debug, you could just add it. I thought this may could cause issues, and usually we do not need to debug from another device. But I am ok with it!

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +46,46 @@ public async Task<FileResult> GetImage(Guid id)
...
+
+        [HttpGet("RandomImageAndInfo", Name = "GetRandomImageAndInfo")]
+        [Produces("application/json")]
+        public async Task<ImageResponse> GetRandomImageAndInfo()
+        {
+            var _settings = new WebClientSettings();
+            var randomImage = await _logic.GetNextAsset() ?? throw new AssetNotFoundException("No asset was found");
+
+            //var randomImageUrl = Url.Action("GetRandomImage", null, new { id = randomImage.Id }, Request.Scheme);
+            var image = await _logic.GetImage(new Guid(randomImage.Id));
```
JW-CH, line 58: Can you add the webhook notification here?

```cs
var notification = new ImageRequestedNotification(new Guid(randomImage.Id));
_ = _logic.SendWebhookNotification(notification);
```

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +46,46 @@ public async Task<FileResult> GetImage(Guid id)
...
         }
+
+        [HttpGet("RandomImageAndInfo", Name = "GetRandomImageAndInfo")]
+        [Produces("application/json")]
+        public async Task<ImageResponse> GetRandomImageAndInfo()
+        {
+            var _settings = new WebClientSettings();
+            var randomImage = await _logic.GetNextAsset() ?? throw new AssetNotFoundException("No asset was found");
+
+            //var randomImageUrl = Url.Action("GetRandomImage", null, new { id = randomImage.Id }, Request.Scheme);
```
JW-CH, line 57: Remove unused code :)

ImmichFrame.WebApi/Controllers/AssetController.cs
```diff
@@ -38,5 +46,46 @@ public async Task<FileResult> GetImage(Guid id)
...
+
+            randomImage.ThumbhashImage!.Position = 0;
+            byte[] byteArray = new byte[randomImage.ThumbhashImage.Length];
+            randomImage.ThumbhashImage.Read(byteArray, 0, byteArray.Length);
+            string thumbHashBase64 = Convert.ToBase64String(byteArray);
+
+            CultureInfo cultureInfo = new CultureInfo(_settings.Language);
+            string photoDate = randomImage.LocalDateTime.ToString(_settings.PhotoDateFormat, cultureInfo) ?? string.Empty;
+
+            var locationParts = _settings.ImageLocationFormat?.Split(',') ?? Array.Empty<string>();
```
JW-CH, line 75: If I want to change the order e.g. `country, city, state` this will not work.

Why not just replace all `city`, `state`, `country` in the `ImageLocationFormat`?

JW-CH: This would also allow more custom `ImageLocationFormat`s like `I was in city, country`

JW-CH (changes requested):

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: done

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: done

ImmichFrame.WebApi/Controllers/AssetController.cs
3rob3: not sure I fully understood. Hopefully this is what you meant. 
- Do you remember why we had the extra split in State? 
- Do we need to implement Country shortening via ISO 3 digit codes like we used to have? I always disable Country or else it comes out as "United States of America", which is ridiculous.

3rob3: Done, also fixed (I think) wrong Settings using injection.

JW-CH (approved):

PR #239, opened 2024-12-22
fix: rearranging and adding information

aviv926:
When I tried to set the DownloadImages option to True it threw errors until I set a folder for DownloadImages ~ ImageCacheFolder.

In addition, I changed the order of the configuration files so that the options related to Cache will be in the appropriate order.


aviv926:
question:

1. Does this ImageCacheFolder only work when there is no access to load images from the server?

3rob3:
I personally think we should remove the settings.json. This was a carryover from the older apps. I prefer using ENV for config since it can easily be changed (and commented out) in Portainer and applied in just a few clicks. What do you think Jan @JW-CH ?

3rob3:
This looks good to me. You good with it @JW-CH ?

README.md
```diff
@@ -119,6 +119,15 @@ Weather is enabled by entering an API key. Get yours free from [OpenWeatherMap][
...
 If you are using Google Calendar, more information can be found [here](https://support.google.com/calendar/answer/37648?hl=en#zippy=%2Cget-your-calendar-view-only).
 
+### Caching
+
+This will download new images into a ImageCache-Folder. Images will be refreshed after the number of days configured in the RenewImagesDuration-Setting. You can enable this features with the DownloadImages-Setting.
+
+To make it work you will need to uncomment `# - PATH/TO/ImageCacheFolder:/app/ImageCache` in your installation file.
+
+> [!NOTE]
+> If you enable the DownloadImages option to True without setting volumes for ImageCacheFolder it will cause an error!
```
JW-CH, line 129: Just to clarify for me, this does not cause an error if you don't set it in your volumes, right? This only throws an error if you uncomment it and do not change the value to a valid folder?

JW-CH: > When I tried to set the DownloadImages option to True it threw errors until I set a folder for DownloadImages ~ ImageCacheFolder.

This should not be the case. Do you have an error message? This is most likely a bug. The "ImageCache"-folder should be created in the application. I like the volume you made :)

PR #252, opened 2025-01-09
add ExcludedPeople

### LINKED ISSUE #248: Add the possibility to exclude people
### Description

The goal is to have an option to exclude people via the configuration

### Use Case

exclude people via UUID and the photos containing the peoples are not selected

### Details

_No response_

### Pre-Submission Checklist

3rob3:
closes #248


ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -38,6 +38,18 @@ private Task<IEnumerable<Guid>> ExcludedAlbumAssets
             }
         }
 
+        private Task<IEnumerable<Guid>>? _excludedPeopleAssets;
+        private Task<IEnumerable<Guid>> ExcludedPeopleAssets
```
JW-CH, line 42: Why do you query for all assets of a person when an asset has the information already, if a Person (People) exist on there?

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -470,9 +512,10 @@ private async Task<bool> LoadRandomAssets()
 
                     if (randomAssets.Any())
                     {
-                        var excludedList = await ExcludedAlbumAssets;
+                        var excludedAlbumList = await ExcludedAlbumAssets;
+                        var excludedPeopleList = await ExcludedPeopleAssets;
 
-                        randomAssets = randomAssets.Where(x => !excludedList.Contains(Guid.Parse(x.Id))).ToList();
+                        randomAssets = randomAssets.Where(x => !excludedAlbumList.Contains(Guid.Parse(x.Id)) && !excludedPeopleList.Contains(Guid.Parse(x.Id))).ToList();
```
JW-CH, line 518: you could access x.People here and have access to a list of person contained on that asset. We might have a Problem when x.People is null because it was not delivered.

JW-CH (changes requested):

PR #253, opened 2025-01-10
Overlay anchor points

### LINKED ISSUE #194: Add different anchor points for the overlay information such as Image Description, Image Location, Current Date, Current Time
### Description

It would be nice to have 
`top-left` 
`top-right`
`bottom-left`
`bottom-right`

as choices on where to display the overlay info of the pictures

### Use Case

Useful when I want to change the default location of the overlay elements

### Details

_No response_

### Pre-Submission Checklist

### LINKED ISSUE #250: time/date/weather in top left corner
### Description

Hello,
I was wondering if it is possible to move the time/date/weather module from bottom left to Top left corner 

Thank you

### Use Case

makes it clean 

### Details

_No response_

### Pre-Submission Checklist

3rob3:
Allow user to set anchor points for clock/weather, image info, and web calendar.
Close #194 
Closes #250


ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -38,6 +38,18 @@ private Task<IEnumerable<Guid>> ExcludedAlbumAssets
             }
         }
 
+        private Task<IEnumerable<Guid>>? _excludedPeopleAssets;
```
JW-CH, line 41: Why is this in this PR? If you create new branches, make sure you branch from main, not from another branch

docker/example.env
```diff
@@ -12,16 +12,20 @@ RenewImagesDuration=30
 Albums=ALBUM1,ALBUM2
 ExcludedAlbums=ALBUM3,ALBUM4
 People=PERSON1,PERSON2
+ExcludedPeople=PERSON1,PERSON2
 Webcalendars=https://calendar.mycalendar.com/basic.ics,webcal://calendar.mycalendar.com/basic.ics
+WebCalendarPosition="top-0 right-0"
```
JW-CH, line 17: Defaults should be something like this: `bottom-[0px]` because bottom-0 is a preset and not every number is supported. For example, `bottom-30` will not work and should be `bottom-[30px]` instead. `bottom-[30%]` is also allowed. Maybe we should document somewhere, that all tailwind syntax is valid. Also maybe this should not be explicitly named `XYPosition` and instead could be custom classes in general for `WebCalendar` but then it probably would make more sense just to have a separate CSS to overwrite this stuff. What do you think @3rob3

JW-CH (changes requested):

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
3rob3: oops, my bad. I switched back to Main first but must not have pulled.

docker/example.env
3rob3: Agree. I originally had a custom position class to convert "BottomLeft" to a CSS compatible string. It got very confusing and I had a lot of trouble getting it to work. Then I thought it would also be nice to adjust from 0 so did it like I did. I couldn't get any value other than 0 to work right so your explanation makes sense.

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
JW-CH: Fixed it for you :) you have to rebase to this branch again.

docker/example.env
JW-CH: https://tailwindcss.com/docs/top-right-bottom-left#arbitrary-values

For more information about the brackets

3rob3: Still doesn't work right, i think because it's parent is not relative? IDK

JW-CH: works for me: `bottom-[50px] right-[80px]`
image

3rob3: hmm, i don't get it. When I put in any value other than zero it gets shoved all the way over to the left or top.

PR #258, opened 2025-01-13
Feat: Add authentication via basic secret

### LINKED ISSUE #255: FR: Super simple authentication
### Description

Right now when I expose immichFrame to the Internet everyone with the link can view the photos.
It would be creat if you could add a super simple login step, could be a token or user/pass combo and also add that to the android app & screensaver. Can be entirely optional.

### Use Case

As a user i'd like my photos to be behind SOME security at least, and not just through obscurity (not knowing the link)

### Details

_No response_

### Pre-Submission Checklist

JW-CH:
Closes #255

### DIFF
--- a/ImmichFrame.Core/Interfaces/IServerSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IServerSettings.cs
@@ -24,5 +24,6 @@
         public string? UnitSystem { get; set; }
         public string Language { get; set; }
         public string? Webhook { get; set; }
+        public string? ImmichFrameAuthenticationSecret { get; set; }
     }
 }
--- a/ImmichFrame.Core/Interfaces/IServerSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IServerSettings.cs
@@ -24,6 +24,6 @@
         public string? UnitSystem { get; set; }
         public string Language { get; set; }
         public string? Webhook { get; set; }
-        public string? ImmichFrameAuthenticationSecret { get; set; }
+        public string? AuthenticationSecret { get; set; }
     }
 }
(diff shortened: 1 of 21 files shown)


3rob3:
Nice.
What do you think about passing the auth in the URL (so clients don't need to be updated)? Something like:
if (Request.Query.TryGetValue("authToken", out var token) && token == _authenticationSecret)
{
    var claims = new[] { new Claim(ClaimTypes.NameIdentifier, "authenticatedUser") };
    var identity = new ClaimsIdentity(claims, Scheme.Name);
    var principal = new ClaimsPrincipal(identity);
    var ticket = new AuthenticationTicket(principal, Scheme.Name);

    return Task.FromResult(AuthenticateResult.Success(ticket));
}

JW-CH:
> Nice. What do you think about passing the auth in the URL (so clients don't need to be updated)? Something like: if (Request.Query.TryGetValue("authToken", out var token) && token == _authenticationSecret) { var claims = new[] { new Claim(ClaimTypes.NameIdentifier, "authenticatedUser") }; var identity = new ClaimsIdentity(claims, Scheme.Name); var principal = new ClaimsPrincipal(identity); var ticket = new AuthenticationTicket(principal, Scheme.Name);
> 
> ```
> return Task.FromResult(AuthenticateResult.Success(ticket));
> ```
> 
> }

I think we need to handle this client side url. The client can take the secret from the URL and pass it in as a auth header. Having the password in the api as a parameter would be a bad pattern in my opiniton. But yea, I agree. Maybe we could make it similar to the ClientIdentifier to have it once via URL and then saved locally and not having to put it again.

JW-CH:
TODO: Update documentation lol

3rob3:
I have tested this pretty well in Android, Android non-WebView, Tauri, and browser. I'm good to merge but should we wait until we are ready to release so documentation doesn't confuse anyone?

PR #259, opened 2025-01-16
Added ShowFavorites option

### LINKED ISSUE #249: Add feature to filter only isFavorite=true
### Description

Currenly, it is possible to add a property to add `isFavorite=true` to the current filter.

### Use Case

On config file or using variables add ImagesFavoriteOnly to select if this field is used to filter the images. Then run the application, only assets tagged with favorite will be shown.

### Details

See the immich api search isFavorite field --> https://immich.app/docs/api/search-smart#isFavorite

### Pre-Submission Checklist

3rob3:
Closes #249


ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -190,6 +190,12 @@ public async Task DeleteAndCreateImmichFrameAlbum()
                 list = list.Union(await GetMemoryAssets());
             }
 
+            if (_settings.ShowFavorites)
+            {
+                assetsAdded = true;
+                list = list.Union(await GetRandomAssets());
```
3rob3, line 196: I'm not sure about this

rjhilgefort:
Much appreciated! This would be great!

3rob3:
> Much appreciated! This would be great! I like the implementation that you can filter albums as well as just show random favorites without an album 🙌

You can actually do both at the same time. It will randomly show an image from favorites and images from Albums.

rjhilgefort:
>> Much appreciated! This would be great! I like the implementation that you can filter albums as well as just show random favorites without an album 🙌

> You can actually do both at the same time. It will randomly show an image from favorites and images from Albums.

That's sweet! I will probably take advantage of that once this merged and shipped.

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
rjhilgefort: Out of curiosity, what's the concern here?

3rob3: > Out of curiosity, what's the concern here?

No concern, it works fine. Just not sure if it's the most efficient way.

JW-CH (approved):

PR #278, opened 2025-02-27
Add invoke to GetWeather timer, simplify time interval

3rob3:
Same issue as found with GetAppointments in #257


immichFrame.Web/src/lib/components/elements/clock.svelte
damongolding, line 1: I’m really surprised Typescript wasn’t moaning about not invoking the function?!

Issue #279, opened 2025-03-01
ImmichFrame no longer working after Immich v1.128 due to manual face tagging


keesjebal:
### Version

v1.0.22.0

### Deployment Method

Android, Web

### Description

Since Immich v1.127 it's possible to manually tag faces. In Immich v1.128.0 they made some [API changes ](https://github.com/immich-app/immich/pull/16364)to handel manually tagged faces. ImmichFrame is not able yet to handle this and is not able to start if there are faces manually tagged. 
I also tried to set 'ShowPeopleDesc' to 'false' but that is also not making a difference.
I only have some manually tagged faces for Favorites. When I set 'ShowFavorites' to false, ImmichFrame is able to start again.

### Reproduction

Do some manual face tagging for Favorites and restart the ImmichFrame container. ImmichFrame will not start. Disable ShowFavorites and it will work again.

### Configuration

```xml
{
  "ImmichServerUrl": "http://172.19.0.1:2283",
  "ApiKey": "***",
  "Margin": "0,0,0,0",
  "Interval": 30,
  "TransitionDuration": 0.5,
  "DownloadImages": false,
  "ShowMemories": true,
  "ShowFavorites": true,
  "ImagesFromDays": 3650,  
  "RenewImagesDuration": 1,
  "RefreshAlbumPeopleInterval": 12,
  "ShowClock": false,
  "ClockFormat": "HH:mm:ss",
  "ShowPhotoDate": true,
  "PhotoDateFormat": "d MMM ''yy",
  "ShowImageDesc": true,
  "ShowPeopleDesc": true,
  "ShowImageLocation": true,
  "ImageLocationFormat": "City",
  "PrimaryColor": "#FF5733",
  "SecondaryColor": "#000000",
  "BaseFontSize": "75px",
  "Style": "transition",
  "UnitSystem": "metric",
  "Language": "en",
  "ImageZoom": true,
  "Layout": "splitview"
}
```

### Logs

```shell
---> System.ArgumentException: Requested value 'manual' was not found.
---> Newtonsoft.Json.JsonSerializationException: Error converting value "manual" to type 'System.Nullable`1[ImmichFrame.Core.Api.SourceType]'. Path '[42].people[0].faces[0].sourceType', line 1, position 101502.
```

### Pre-Submission Checklist

- [x] This is a bug report and not a feature request

- [x] I have provided all of the required information to reproduce the bug (config, logs, etc.)

- [x] I have checked for related issues and checked the documentation

[...]

PR #283, opened 2025-03-06
Add the 'manual' option to the sourceType enum

dracoling:
This fixes the API schema error in #279 by adding the new enum value added in the recent update.


3rob3:
Nice, thanks! Did you verify it locally?

dracoling:
Yes, this is working currently with my immich v1.129.

3rob3:
Closes #279

PR #287, opened 2025-03-19
Fix: Height on mobile devices by switching from h-screen to h-dvh

jwlarocque:
On mobile devices with a dynamic viewport size ("URL bar" can be shown/hidden), `h-screen` sets an element's height to the maximum possible viewport size.  This results in the top or bottom of the application being obscured by the URL bar.  
`h-dvh` instead uses the `dvh`  unit to set the element height to the current actual viewport height.  `dvh` is supported in all browsers, most since late 2022, except IE 11.

Only tested in Firefox and Chromium on Linux, Firefox and Chrome on Android, and Safari on iOS, as I am not set up to build and test the native clients.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -5,17 +5,11 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <script src="https://cdn.tailwindcss.com?compatibility"></script>
-    <style>
-        #full_section {
-            height: 100vh;
-            height: 100dvh;
-        }
-    </style>
 </head>
 
 <body class="bg-gray-100 overflow-hidden">
     <!-- Full Section with Flexbox to Ensure Everything Fits -->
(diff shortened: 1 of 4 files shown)


3rob3:
Did you happen to test this on iOS installed as a PWA?

jwlarocque:
Good idea, I just tested that and it looks as I'd expect (on main the progress bar was cut off, with the change it fits to the visible view height).

I also went looking for devices (I own) without dvh support.  Couldn't find any but I figure there are probably some people using very old devices as frames so I added fallback to vh.  Unfortunately I don't think this is possible with tailwind so it complicates the CSS slightly.

jwlarocque:
Makes sense; should I roll back that commit?

3rob3:
> Makes sense; should I roll back that commit?

Yeah, I think so. The fallback one right?

3rob3:
looks good to me.

PR #288, opened 2025-03-19
Improve responsiveness when user manually requests prev/next image

jwlarocque:
Manually paging through images currently feels somewhat sluggish due to a combination of transition time (which is of course desirable for non-manual transitions) and long image load times when connected to a remote api server.

This PR includes two proposed changes in order to improve responsiveness:
* set transition duration to zero when the user manually requests the previous or next image (and then set it back to the configured value at the next automatic transition)
* set cache-control headers to store images in the browser cache for up to one week

I'd be happy to rework these if another approach would be preferable, please let me know.

There's one thing which I did _not_ attempt to implement, but which I think could be quite beneficial: pre-loading the next/"on-deck" image.  I didn't see an easy way to incorporate this into the current structure of the code but I think it could further improve perceived UI responsiveness and also improve the consistency of automatic slide advances.

### DIFF
--- a/ImmichFrame.WebApi/Controllers/AssetController.cs
+++ b/ImmichFrame.WebApi/Controllers/AssetController.cs
@@ -6,7 +6,6 @@ using ImmichFrame.Core.Logic;
 using ImmichFrame.WebApi.Models;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.Net.Http.Headers;
 
 namespace ImmichFrame.WebApi.Controllers
 {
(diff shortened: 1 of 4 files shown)


3rob3:
Thanks for the contribution!
I like the instant transition change, great idea.
I'm uncertain about the cache change. We have cache purposely disabled in the WebView apps and not sure how this change would effect that (if at all). It also could lead to inconsistent performance, depending on whether the next asset is cached or not (some images loading quicker than others). Would you mind splitting the cache change into a separate PR? I have no problem merging the instant transition changes but would prefer @JW-CH reviews the cache one, and he is unavailable for another month or two.

JW-CH:
Month or two? Back in a few days :P

3rob3:
> Month or two? Back in a few days :P

thought you said 3 months, must have been weeks :). oops.

3rob3:
Just wanted to make sure you saw this comment:
Would you mind splitting the cache change into a separate PR?

jwlarocque:
Distracted myself messing around with some other load changes...  
I've taken out the cache headers; I'll try a few more things with pre-loading images and submit a new PR for one or the other once I'm satisfied/give up.

3rob3:
@JW-CH , looks good to me, what do you think? Seems like it should be simpler to do this, but I tried and was unsuccessful.

JW-CH (approved):
LGTM

PR #293, opened 2025-03-25
Fix: Add animation fill on image zoom

jwlarocque:
The image zoom effect "jumps" at the end of its animation back to its un-transformed state.  Normally the slideshow will have moved on before this happens, due to the 2s extra animation length, but if it is paused or the next image takes a particularly long time to load the jump will be visible to the user.

This PR adds a forwards animation fill to all the image zoom animations which should fix the issue.  (The pure `zoom-out` is currently unaffected since `scale(1)` is an identity transform, but I think it's probably good to put the fill in now in case of future changes to the animation.)

### DIFF
--- a/immichFrame.Web/src/lib/components/elements/image.svelte
+++ b/immichFrame.Web/src/lib/components/elements/image.svelte
@@ -168,16 +168,16 @@
 
 <style>
 	.zoom-in {
-		animation: zoom-in var(--interval) ease-out normal;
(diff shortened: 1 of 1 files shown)


3rob3:
I have noticed this as well. I had a lot of issues with this in the non-WebView Android and Apple TV implementations also. On mobile so can’t review now but will hopefully tomorrow.

JW-CH (approved):
Once again, thank you for these changes. I was looking for this but could not figure it out 🎉

PR #297, opened 2025-03-28
Feat: Pre-fetch and cache images

jwlarocque:
Okay, this might need some additional work but I want to get your feedback on the approach (I've reimplemented it a couple of times, this is the best I could find).

This PR would:
* move image loading to home-page
* create a dictionary of loadImage promises, which is updated when the slideshow advances
* load data into imagesState by awaiting those promises
* set cache headers to store images for one week (this is currently relied upon for loading previous images, but we could keep those in JS as well with this approach if we wanted to)

Some things I'm not sure about:
* PRELOAD_IMAGES (the distance ahead in the backlog to fetch) could be configurable?  If it was it should probably also be capped so people don't set it to some huge number.
* It might be worthwhile to construct some kind of intermediate component to handle image loading?  home-page is getting kind of crowded.
* 3rob3 mentioned that caching is currently intentionally disabled for the native/webview clients - I was wondering what the reason for this is.  If it's not always undesirable, maybe it could also be user configurable?

### DIFF
--- a/immichFrame.Web/src/lib/components/home-page/home-page.svelte
+++ b/immichFrame.Web/src/lib/components/home-page/home-page.svelte
@@ -23,7 +23,7 @@
 
 	let displayingAssets: api.AssetResponseDto[] = $state() as api.AssetResponseDto[];
 
-	const { restartProgress, stopProgress } = slideshowStore;
+	const { restartProgress, stopProgress, instantTransition } = slideshowStore;
 
 	let progressBarStatus: ProgressBarStatus = $state(ProgressBarStatus.Playing);
 	let progressBar: ProgressBar = $state() as ProgressBar;
@@ -78,8 +78,9 @@
 		}
 	}
 
-	const handleDone = async (previous: boolean = false) => {
+	const handleDone = async (previous: boolean = false, instant: boolean = false) => {
 		progressBar.restart(false);
+		$instantTransition = instant;
 		if (previous) await getPreviousAssets();
 		else await getNextAssets();
 		progressBar.play();
@@ -224,10 +225,10 @@
 
 		<OverlayControls
 			on:next={async () => {
-				await handleDone();
+				await handleDone(false, true);
 			}}
 			on:back={async () => {
-				await handleDone(true);
+				await handleDone(true, true);
 			}}
 			on:pause={async () => {
 				if (progressBarStatus == ProgressBarStatus.Paused) {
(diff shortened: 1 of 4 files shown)


JW-CH:
Sorry for the late reply, I'll take a closer look at this in the next few days. Thank you for the contribution!

immichFrame.Web/src/lib/components/home-page/home-page.svelte
```diff
@@ -16,20 +16,40 @@
...
+		error: boolean;
+		loaded: boolean;
+		split: boolean;
+		hasBday: boolean;
+	}
+
 	api.init();
 
+	// TODO: make this configurable?
+	const PRELOAD_IMAGES = 5;
```
JW-CH, line 30: I think this should be configurable? Devices with low memory could run out of memory maybe?

immichFrame.Web/src/lib/components/home-page/home-page.svelte
```diff
@@ -158,6 +208,62 @@
...
+				images: [],
+				error: true,
+				loaded: false,
+				split: false,
+				hasBday: false
+			};
+		}
+	}
+
+	async function loadImage(a: api.AssetResponseDto) {
```
JW-CH, line 254: Could you give 'a' a proper name :)

immichFrame.Web/src/lib/components/home-page/home-page.svelte
```diff
@@ -158,6 +208,62 @@
...
+		let req = await api.getImage(a.id, { clientIdentifier: $clientIdentifierStore });
+
+		if (req.status != 200) {
+			return ['', a] as [string, api.AssetResponseDto];
```
JW-CH, line 258: If the response is not ok for some reason, an error will be displayed. Some kind of 'retry' would be nice.

Maybe if the image url is `''`, it should retry loading them. You can play around with the chrome network settings (setting it offline). I am also ok doing this with another PR.

JW-CH (changes requested):

JW-CH:
Perfect change, love it. Thank you!

3rob3:
> Perfect change, love it. Thank you!

What's the status of this? Any reason it is not merged?

JW-CH:
> > Perfect change, love it. Thank you!
> 
> What's the status of this? Any reason it is not merged?

Had some nice to have changes. Will will merge this and change them myself.

PR #306, opened 2025-04-13
Add version, version logging, and version endpoint

3rob3:
There is currently no way to see backend version number. This will add it to log and also expose an endpoint that can be checked via a client if ever required.

### DIFF
--- a/ImmichFrame.WebApi/Controllers/ConfigController.cs
+++ b/ImmichFrame.WebApi/Controllers/ConfigController.cs
@@ -23,5 +23,15 @@ namespace ImmichFrame.WebApi.Controllers
         {
             return (WebClientSettings)_settings;
         }
+        [HttpGet("version")]
+        public IActionResult GetVersion()
+        {
+            var version = System.Reflection.Assembly
+                .GetExecutingAssembly()
+                .GetName()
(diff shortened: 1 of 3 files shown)


ImmichFrame.WebApi/Controllers/ConfigController.cs
```diff
@@ -23,5 +23,15 @@ public WebClientSettings GetConfig(string clientIdentifier = "")
         {
             return (WebClientSettings)_settings;
         }
+        [HttpGet("GetVersion")]
+        public IActionResult GetVersion()
```
JW-CH, line 27: Can you return a string instead of a ActionResult?

JW-CH (changes requested):

ImmichFrame.WebApi/Controllers/ConfigController.cs
3rob3: Sure, makes more sense. Done

JW-CH:
Probably a TODO: Add some kind of version file to not have to set this multiple times in code.

3rob3:
it is kind of annoying, but you should only need to set it in "ImmichFrame.WebApi.csproj". I was trying to think of a way that pushing a tag also sets this value.

PR #307, opened 2025-04-14
don't show archived assets

### LINKED ISSUE #304: Don't show archived assets
### Description

Option (or hardcode) to not show archived assets even if they belong to an album that's picked to display.

immich-kiosk has this option: `show_archived: false # Allow assets marked as archived to be displayed`

### Use Case

Album management in immich is still very primitive, so we're forced to use the Archive for many things.
See: https://github.com/immich-app/immich/discussions/17069

### Details

_No response_

### Pre-Submission Checklist

3rob3:
Closes #304 

I'm not real confident on this one, feel free to reject.
Also, why do we have "WithPeople = true" in LoadRandomAssets? That doesn't seem right.


VelocityRa:
Why closed? 👀

3rob3:
This PR conflicted with another, more complex PR. Once that one is merged I will redo these changes.

> Why closed? 👀

PR #308, opened 2025-04-14
Feat: Rating (Stars) Filter

### LINKED ISSUE #303: Add Feature to filter by star rating
### Description

Similar to #249 but for a star rating, which is a bit more granular - nice to have feature

### Use Case

e.g rating=3 would filter to all pictures which has 3 or more stars. as far I know -1 is all without stars. 

### Details

https://immich.app/docs/api/search-smart/

### Pre-Submission Checklist

JW-CH:
Closes #303


JW-CH:
@3rob3 FYI: The immich-api filtering only returns assets where the rating (stars) is the exact same as the one you are looking for.

Meaning:
Searching for assets with a rating of 4 will only return ones with a 4 and none with a higher rating. Maybe this is a bug? I implemented it in the same way for now.

Bierchermuesli:
very nice. Many thanks for this quick implementation. Code looks nice and simpler than I thought 

Maybe a Bug or just not foreseen. maybe a dedicated operator `eq, ge, le...` would make sens in their api?

docker/Settings.example.json
```diff
@@ -7,6 +7,7 @@
   "DownloadImages": false,
   "ShowMemories": false,
   "ShowFavorites": false,
+  "Rating": 5,
```
3rob3, line 10: Need this in example.env also

3rob3 (changes requested):
Looks good other than my one comment (example.env)

JW-CH:
> very nice. Many thanks for this quick implementation. Code looks nice and simpler than I thought
> 
> Maybe a Bug or just not foreseen. maybe a dedicated operator `eq, ge, le...` would make sens in their api?

I'd say so, a dedicated operator would be great. I'll ask when I find a minute.

JW-CH:
Created a discussion (feature request) on the immich repo: https://github.com/immich-app/immich/discussions/17616

PR #312, opened 2025-04-21
isArchived setting and implementation

### LINKED ISSUE #304: Don't show archived assets

3rob3:
closes #304


ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -344,7 +344,10 @@ private async Task<IEnumerable<AssetResponseDto>> GetPeopleAssets()
                                 WithExif = true,
                                 WithPeople = true
                             };
-
+                            if (!_settings.ShowArchived)
+                            {
+                                metadataBody.IsArchived = false;
```
JW-CH, line 349: Why not ` metadataBody.IsArchived = _settings.ShowArchived`?

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
```diff
@@ -460,6 +463,11 @@ private async Task<bool> LoadRandomAssets()
                         WithExif = true,
                         WithPeople = true,
                     };
+                    
+                    if (!_settings.ShowArchived)
+                    {
+                        searchBody.IsArchived = false;
```
JW-CH, line 469: Same as other comment.

ImmichFrame.Core/Logic/ImmichFrameLogic.cs
3rob3: Because if setting is true, you will ONLY get back archived photos (instead of both regular and archived). Therefore you need to only include it if it's false. It's confusing, and I also expect this could change on Immich side. That's why I originally just hard coded it to false with no setting.

PR #317, opened 2025-04-24
Feat: Rework Backend, decouple

JW-CH:
- Decouple Weather Logic
- Decouple Calendar Logic
- New Implementation of the ImmichFrame backend
- Add Cache class
- Sanitize user identifier in log/webhook

### DIFF
--- a/ImmichFrame.Core/Interfaces/IImmichFrameLogic.cs
+++ b/ImmichFrame.Core/Interfaces/IImmichFrameLogic.cs
@@ -1,12 +1,11 @@
 ﻿using ImmichFrame.Core.Api;
-using OpenWeatherMap;
 
 namespace ImmichFrame.Core.Interfaces
 {
     public interface IImmichFrameLogic
     {
         public Task<AssetResponseDto?> GetNextAsset();
-        public Task<List<AssetResponseDto>> GetAssets();
+        public Task<IEnumerable<AssetResponseDto>> GetAssets();
         public Task<(string fileName, string ContentType, Stream fileStream)> GetImage(Guid id);
         // public Task AddAssetToAlbum(AssetResponseDto assetToAdd);
         // public Task DeleteAndCreateImmichFrameAlbum();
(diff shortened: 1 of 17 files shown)


ImmichFrame.Core/Logic/OptimizedImmichFrameLogic.cs
```diff
@@ -21,19 +21,314 @@ public void Dispose()
         _httpClient.Dispose();
     }
 
-    public Task<List<AssetResponseDto>> GetAssets()
+    private Queue<AssetResponseDto> _assetQueue = new();
+    private bool _isReloadingAssets = false;
+
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+    public async Task<AssetResponseDto?> GetNextAsset()
```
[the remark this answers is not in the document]
JW-CH: Yes, thank you :)

JW-CH:
@3rob3 I added the [action] to the webapi-endpoints.

This will break the app
You need to call
- `/api/Asset/GetRandomImageAndInfo` instead of `/api/Asset/RandomImageAndInfo`
- `api/Config/GetConfig` instead of `api/Config`
- `api/Weather/GetWeather` instead of `api/Weather`

Is this a NO-GO?

ImmichFrame.Core/Interfaces/IImmichFrameLogic.cs
```diff
@@ -1,18 +1,15 @@
...
+        public Task<IEnumerable<AssetResponseDto>> GetAssets();
+        public Task<AssetResponseDto> GetAssetInfoById(Guid assetId);
         public Task<(string fileName, string ContentType, Stream fileStream)> GetImage(Guid id);
-        public Task AddAssetToAlbum(AssetResponseDto assetToAdd);
-        public Task DeleteAndCreateImmichFrameAlbum();
-        public Task<List<IAppointment>> GetAppointments();
-        public Task<IWeather?> GetWeather();
-        public Task<IWeather?> GetWeather(double latitude, double longitude, OpenWeatherMapOptions Options);
+        // public Task AddAssetToAlbum(AssetResponseDto assetToAdd);
+        // public Task DeleteAndCreateImmichFrameAlbum();
```
3rob3, line 12: Not that it is necessary for this PR, but I say we just remove anything to do with this now obsolete feature. I never liked my original implementation anyway.

JW-CH: Sounds good to me

3rob3 (approved):
LGTM. I tested with Android and Android widgets also.

PR #333, opened 2025-05-05
ImageFill

### LINKED ISSUE #300: Request: Enhanced zoom settings (possibility to stretch to frame size from the start)
### Description

I love the transition effect between photos and also the gradual zooming, but I would like the picture to fill the frame from the start, without any blurry or black bars.

### Use Case

The original frame software offers a very slight gradual zoom, the pictures already fill the screen when the zoom process starts and are just zoomed into ever so slightly, which creates a very elegant effect and makes the whole frame less boring to look at. I would love to be able to have this implemented alongside the great zoom feature that is already there. My photos have a very high resolution and even if they fill the frame at the beginning, they could still easily be zoomed into. Having blurry bars might be nice for smartphone pictures, but I have a digital camera that produces high-quality pictures and I feel they would look much nicer that way.
If this is considered to be a duplicate of [Issue 267](https://github.com/immichFrame/ImmichFrame/issues/267), simply close my request.

### Details

_No response_

### Pre-Submission Checklist

3rob3:
Implement ImageFill setting (replaces unused ImageStretch). If true will fill available space without distortion.

This will need to be implemented on Android Frameo and Apple TV separately.

Closes #300

### DIFF
--- a/immichFrame.Web/src/lib/components/elements/asset-info.svelte
+++ b/immichFrame.Web/src/lib/components/elements/asset-info.svelte
@@ -53,7 +53,7 @@
 	let availablePeople = $derived(asset.people?.filter((x) => x.name));
 </script>
 
-{#if showPhotoDate || showPhotoDate || showImageDesc || showPeopleDesc}
+{#if showPhotoDate || showLocation || showImageDesc || showPeopleDesc}
 	<div
 		id="imageinfo"
 		class="immichframe_image_metadata absolute bottom-0 right-0 z-100 text-primary p-1 text-right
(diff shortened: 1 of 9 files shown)


immichFrame.Web/src/lib/components/elements/asset-info.svelte
```diff
@@ -53,7 +53,7 @@
 	let availablePeople = $derived(asset.people?.filter((x) => x.name));
 </script>
 
-{#if showPhotoDate || showPhotoDate || showImageDesc || showPeopleDesc}
+{#if showPhotoDate || showLocation || showImageDesc || showPeopleDesc}
```
JW-CH, line 56: Thanks for this copy and paste fix :P

JW-CH (approved):

PR #339, opened 2025-05-07
Feat: Fetch asset metadata for images without people

JW-CH:
immich does not return people data when using the filtering for albums. This change (pre)fetches this data. This will result in e.g. people being displayed if you have an album filter active.


3rob3:
👍

PR #346, opened 2025-05-09
Clock update fix

### LINKED ISSUE #345: Clock display only updates when image refreshes
### Version

v1.0.24.0 and Client:V39

### Deployment Method

SSA Photo Frame (similar to frameo I suppose)

### Description

I have set the refresh interval to 3600 (1 hour), and I noticed that the time (clock) on the frame only updates when the image is refreshed — in this case, once per hour. When I changed the interval to 2 minutes, the time updates every 2 minutes accordingly.

It seems the time is tied to the image refresh cycle.
Is it possible to have the time update every minute independently of the image refresh? Ideally, the clock would stay accurate without requiring a full image reload.

### Reproduction

Change the image refresh interval, and notice that the clock update also follows that same interval

### Expectations

Clock updates should be frequent - at least once a minute

### Configuration

```xml

```

### Logs

```shell
I dont see any particular logs
```

### Pre-Submission Checklist

- [x] This is a bug report and not a feature request

- [x] I have provided all of the required information to reproduce the bug (config, logs, etc.)

3rob3:
writeable for time
Closes #345

### DIFF
--- a/immichFrame.Web/src/lib/components/elements/clock.svelte
+++ b/immichFrame.Web/src/lib/components/elements/clock.svelte
@@ -1,40 +1,50 @@
 <script lang="ts">
 	import * as api from '$lib/index';
-	import { onMount } from 'svelte';
+	import { onMount, onDestroy } from 'svelte';
 	import { format } from 'date-fns';
 	import * as locale from 'date-fns/locale';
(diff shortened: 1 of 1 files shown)


JW-CH:
#350 uses the rune mode of svelte, closing this

PR #356, opened 2025-05-15
Feat: Display more details

### LINKED ISSUE #242: Add a button to show QR code and Exif info
### Description

ImmichKiosk has an option to add a QR code. It would also be nice to show the Exif info. (lens info)

### Use Case

Provides more info to users and guests. 

### Details

The user will press a button to show a QR code and exif info (mm, apeture etc)

### Pre-Submission Checklist

JW-CH:
- Basic implementation of a 'more details'/'info' page
- QR-Code takes the base URL from the existing `ImmichServerUrl`
- Image overlay hides the action overlay (<, | |, >)
- Can be closed via the 'x' top right
- Shortcut 'i' to open/close the info overlay

To-do: Implement in client? (Shortcut and for android TV remotes) (cc @3rob3 )

Closes #242

### DIFF
--- a/ImmichFrame.Core/Interfaces/IClientSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IClientSettings.cs
@@ -7,6 +7,7 @@
 
     public interface IClientSettings
     {
+        public string ImmichServerUrl { get; set; }
         public string Margin { get; set; }
         public int Interval { get; set; }
         public double TransitionDuration { get; set; }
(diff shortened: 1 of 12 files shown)


3rob3:
> > Do you think we should trigger a Pause when the info is being shown?
> 
> Yes. To scan the QR etc. We could resume play on closing. (Hit the shortcut 'i' does that)

Oh, your doing that already. I was suggesting we pause because i didn't see it.

PR #361, opened 2025-05-20
Fix: Frameo WebView Fixes

3rob3:
Not working/ To do:
Splitview left photo info still not showing correctly. On transition the photo info change is pretty abrupt/harsh.

### DIFF
--- a/immichFrame.Web/src/lib/components/elements/asset-info.svelte
+++ b/immichFrame.Web/src/lib/components/elements/asset-info.svelte
@@ -68,7 +68,7 @@
 {#if showPhotoDate || showLocation || showImageDesc || showPeopleDesc || showAlbumName}
 	<div
 		id="imageinfo"
-		class="immichframe_image_metadata absolute bottom-0 right-0 z-100 text-primary p-1 text-right
+		class="immichframe_image_metadata fixed bottom-0 right-0 z-100 text-primary p-1 text-right
 		{$configStore.style == 'solid' ? 'bg-secondary rounded-tl-2xl' : ''}
 		{$configStore.style == 'transition' ? 'bg-gradient-to-l from-secondary from-0% pl-10' : ''}
 		{$configStore.style == 'blur' ? 'backdrop-blur-lg rounded-tl-2xl' : ''}	"
--- a/immichFrame.Web/src/lib/components/elements/asset-info.svelte
+++ b/immichFrame.Web/src/lib/components/elements/asset-info.svelte
@@ -68,7 +68,7 @@
 {#if showPhotoDate || showLocation || showImageDesc || showPeopleDesc || showAlbumName}
 	<div
 		id="imageinfo"
-		class="immichframe_image_metadata fixed bottom-0 right-0 z-100 text-primary p-1 text-right
+		class="immichframe_image_metadata absolute bottom-0 right-0 z-100 text-primary p-1 text-right
 		{$configStore.style == 'solid' ? 'bg-secondary rounded-tl-2xl' : ''}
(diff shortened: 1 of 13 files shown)


immichFrame.Web/src/lib/components/elements/asset-info.svelte
```diff
@@ -68,7 +68,7 @@
 {#if showPhotoDate || showLocation || showImageDesc || showPeopleDesc || showAlbumName}
 	<div
 		id="imageinfo"
-		class="immichframe_image_metadata absolute bottom-0 right-0 z-100 text-primary p-1 text-right
+		class="immichframe_image_metadata fixed bottom-0 right-0 z-100 text-primary p-1 text-right
```
JW-CH, line 71: This needs to be `absolute`! This will hide the first images assetinfo when in split mode!

immichFrame.Web/src/lib/components/elements/image-component.svelte
```diff
@@ -70,7 +70,7 @@
 	<ErrorElement />
 {:else if loaded}
 	{#key images}
-		<div class="grid absolute h-dvh w-screen" transition:blur={{ duration: transitionDuration }}>
+		<div class="grid fixed h-dvh w-screen" transition:blur={{ duration: transitionDuration }}>
```
JW-CH, line 73: this should not be needed and can be removed.

JW-CH (changes requested):

immichFrame.Web/src/lib/components/elements/asset-info.svelte
3rob3: You get no photo info at all with absolute. As mentioned above, fixed is not correct either because first images have no date when in splitview.

JW-CH: `absolute` is the correct class to use here. No way to get around that as of now. What is not working on the frameo then? What version of web view are you using? I want to test that on an emulator or something.

3rob3: Download Chrome 106 and you can see the issues. In this branch weather/clock is fine, but photo info/people usually won't show or is incomplete. Also overlay controls are in the wrong places (I have seen this before on other devices). 

On demo server or main clock/weather won't show, so I think at least that is fixed with this branch.

3rob3: What is really strange is if you open developer tools in chrome to inspect, photo info suddenly shows. Seems like this all points to an issue with the outer control, and these elements are being placed off screen

immichFrame.Web/src/lib/components/elements/image-component.svelte
JW-CH: by 'this should not be needed', I meant the absolute/fixed can be removed. Please verify

immichFrame.Web/src/lib/components/elements/image-component.svelte
```diff
@@ -70,7 +70,7 @@
 	<ErrorElement />
 {:else if loaded}
 	{#key images}
-		<div class="grid fixed h-dvh w-screen" transition:blur={{ duration: transitionDuration }}>
+		<div class="grid absolute h-dvh w-screen" transition:blur={{ duration: transitionDuration }}>
```
JW-CH: Why did you add this again?

3rob3: I didn't. It is not one of the "files changed" anymore

JW-CH:
I made some more changes that fix the issues with older Chrome version. Hope you are happy with them 🚀

JW-CH (approved):

3rob3:
working great for me!

PR #374, opened 2025-05-29
Feat: Multiple Immich account support

jpg0:
Added ability to configure multiple immich accounts to pull photos from. Each account can have all configuration options applying to an account (like filtering etc). Randomly selects from accounts provided, each account getting equal chance.

Configuration works like:
```
Account1.ApiKey=...
Account2.ApiKey=...
```
Also allows omitting `AccountX.` prefix which defaults to first account to maintain backwards compatibility.

For looking up details of assets via guid, it currently handles this by caching (for a minute) the guids by account. This means that they can be looked up within a minute of being handed out, otherwise the lookup will fail. I assume this should work for most use-cases, but can add a fallback of iterating through accounts or something if required.


jpg0:
I've added a second commit which feeds images from accounts proportionally based on the total number of images per account (refreshed daily). Ideally it would sum the total amount of images searchable after filters etc, but this is probably a reasonable improvement.

JW-CH:
Love the change so far, skipped through the code for a bit. Would you mind also updating the docs? Thank you for your contribution.

jpg0:
Sure, no problem! I'll get to it over the next few days.

ImmichFrame.Core/Logic/TotalAccountImagesSelectionStrategy.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    private async Task<IList<double>> GetProportions(IList<IImmichFrameLogic> accounts)
+    {
+        var (totals, sum) = await GetWeights(accounts);
+        return totals.Select(t => (double)t / sum).ToList();
+    }
+
+    private Task<int> GetTotalForAccount(IImmichFrameLogic account)
+    {
+        return _accountToTotal.GetOrCreateAsync(account, async entry => (await account.GetAssetStats()).Images,
```
JW-CH, line 57: GetAssetStats ignores the actually filtered assets. Lets say Account1 has 100k images, but in the settings I only show the memories, so the ones with the day/month the same as today. For the second account, I show everything, lets say 10 images. With the later takeproportional, there won't be any assets from the second account and only the first accounts 10 images will be displayed. You probably should look on the actual filtered assets.

JW-CH (changes requested):
Looking good, just some small changes.

ImmichFrame.WebApi/Program.cs
```diff
@@ -77,22 +78,45 @@ _ _ __ ___  _ __ ___  _  ___| |__ | |_ _ __ __ _ _ __ ___   ___
...
+builder.Services.AddSingleton<IAccountSelectionStrategy>(srv =>
+{
+    var settings = srv.GetRequiredService<IServerSettings>();
+
+    if (settings.Accounts.Count == 0)
+    {
+        throw new SettingsNotValidException("No immich accounts configured");
+    }
+
+    return ActivatorUtilities.CreateInstance<TotalAccountImagesSelectionStrategy>(srv);
```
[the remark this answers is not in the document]
jpg0: The way I understand it is that using `ActivatorUtilities.CreateInstance` delegates the construction of the instance to the DI framework, rather than creating it yourself. This is more in-line with how you should use DI frameworks in letting them handle as much as possible. For example, doing it this way means that if you were to add a logger to the constuctor of `TotalAccountImagesSelectionStrategy`, then you don't need to update this class too. Also if you wanted to swap out the concrete class type of another class which is passed an interface, you can do this just by registering a different type with DI (which will then handle passing the other type in), rather than needing to explicitly then change what you are passing in. Also, whilst I'm not an expert in this DI system, usually there are testing and mocking frameworks which rely on being able to intercept type registration and construction to allow you to switch out components/instances for testing. So for this specific case it probably doesn't matter much, but I believe that it's a better pattern as the codebase grows, plus for some cases (such as `MultiImmichFrameLogicDelegate`, where some of the constructor params are factory functions) it's much cleaner to do this way. I tried not to change existing code, but actually that whole DI section can be optimised significantly, I'll submit   another PR for it.

ImmichFrame.Core/Logic/TotalAccountImagesSelectionStrategy.cs
jpg0: Absolutely! Sorry I wasn't clear on how I was tackling this, but my intention was to implement this whole feature in steps each of which are releasable (and smaller to review):
- first: just randomly pick between accounts, so they effectively each get a 50% chance to show a photo (this was effectively the first commit of the PR)
- second: switch to a slightly more complex option where it looks at the total number of photos per account, and pull out the logic to do this into distinct strategy classes (with the above becoming the `SimpleAccountSelectionStrategy`)
- third: implement a third strategy which correctly counts the total number of filtered assets
This PR I originally intended to only get to step 1, but accidentally committed and pushed step 2 changes into the same PR, so just thought that I'd skip step one! I certainly plan to do step three - up to you if you want it as part of this PR or a separate one.

jpg0: I've taken a quick look at the Immich API, and I can't see a way to get search result totals without pulling all assets for the search. So my suggestion is that we get in the current changes in this PR (inc. doc changes that I'll add shortly), then I'll open a PR against Immich to return search result totals, then another PR here to add the correct filtered asset totals.

jpg0: Actually... I'm confused. I had assumed that all the assets were fetched on demand whenever one (or some) were needed, by using `searchRandom`. It looks like this is the case when favourites/memories/albums/people are not defined, but if they are then they are all loaded into memory? I guess this is kind-of the same problem as we have across accounts: that you don't know how many photos there in favourites versus another category, so they are all loaded, the selected from the entire loaded collection. If I can make a change to return totals from the Immich API, then presumably we can also modify `OptimizedImmichFrameLogic` to use a similar pattern and use `selectRandom` based on the totals observed from each cateogy (and not load all into memory).

JW-CH: Yes, all the needed `AssetResponseDto`'s will get fetched on startup and cached, the ImmichFrame API will then pick a set amount of `AssetResponseDto` for the client. The immich API does currently not fit our needs with their `searchRandom`-Endpoint since we can configure more stuff. Do you see a way to not load everything into memory?

jpg0: What I'd like to so is ensure that each query response in Immich returns the total number of available results (either in another field, or another similar endpoint). Then for whatever queries we make here, we only cache the totals (we can set the size to 1 and discard the actual asset). If we have all the totals, then we can generate a random number to choose from which 'pool' (memories, people, whatever) based on how many in each pool, assuming that a single `searchRandom` call can query that pool. I see that it can query some (like photos with a person in, favourites), but cannot others (like memories, photos in album). But it seems reasonable to add albums to `searchRandom`, and possible memories too, although we could use a mixed approach until then.

jpg0:
Added documentation

ImmichFrame.WebApi/Program.cs
```diff
@@ -66,9 +67,16 @@ _ _ __ ___  _ __ ___  _  ___| |__ | |_ _ __ __ _ _ __ ___   ___
 });
 
 builder.Services.AddSingleton<IServerSettings, ServerSettings>();
+builder.Services.AddSingleton<IImmichFrameSettings>(srv => srv.GetRequiredService<IServerSettings>().ImmichFrameSettings);
```
JW-CH: I already commented on the other PR (#377), settings as json/yaml file is supported rn. This would break this functionality.

ImmichFrame.WebApi/Models/ServerSettings.cs
```diff
@@ -1,51 +1,94 @@
...
+            public int? Rating { get; set; }
+
+            public string ImmichFrameAlbumName { get; set; } = string.Empty;
+        }
+
+        public IImmichFrameSettings ImmichFrameSettings { get; set; } = new ServerFrameSettings();
+        public IList<IImmichAccountSettings> Accounts { get; set; } = new List<IImmichAccountSettings>();
+
+
+        private Regex _accountSettingRegex = new(@"^Account(\d)\.(\w+)$");
```
JW-CH: This is nitpicking, sorry. What do you think about using named regex groups?

```suggestion
        private Regex _accountSettingRegex = new(@"^Account(?<id>\d)\.(?<property>\w+)$");
```

ImmichFrame.WebApi/Models/ServerSettings.cs
```diff
@@ -1,51 +1,94 @@
...
-                    var propertyInfo = typeof(ServerSettings).GetProperty(key.ToString() ?? string.Empty);
+                    if (key == null || sKey == null) continue;
 
-                    if (propertyInfo != null)
+                    var matches = _accountSettingRegex.Matches(sKey);
+
+                    if (matches.Any())
                     {
-                        this.SetValue(propertyInfo, env[key]?.ToString() ?? string.Empty);
+                        var idx = Int32.Parse(matches[0].Groups[1].Value);
```
JW-CH: Use mentioned named group.

```suggestion
                        var idx = Int32.Parse(matches[0].Groups["id"].Value);
```

ImmichFrame.WebApi/Models/ServerSettings.cs
```diff
@@ -1,51 +1,94 @@
...
+                        var strKey = matches[0].Groups[2].Value;
```
JW-CH: Use mentioned named group.
```suggestion
                        var strKey = matches[0].Groups["property"].Value;
```

ImmichFrame.Core/Logic/OptimizedImmichFrameLogic.cs
```diff
@@ -29,54 +32,39 @@ public void Dispose()
         _httpClient.Dispose();
     }
 
-    private Channel<AssetResponseDto> _assetQueue = Channel.CreateUnbounded<AssetResponseDto>();
```
JW-CH: Also a merge issue?

ImmichFrame.WebApi/Program.cs
jpg0: Submitted #381 which will fix this loading

ImmichFrame.WebApi/Models/ServerSettings.cs
jpg0: fine by me; I'll update.

ImmichFrame.Core/Logic/OptimizedImmichFrameLogic.cs
jpg0: oh, yes. I'll fix that.

jpg0:
I have rebased this onto the fixed config loading, and addressed the points above. However, I have not yet implemented multiple accounts via JSON config. I can do so, but I need more direction on where you want the config structure to go:

Currently the structure is flat - a simple list of keys. This maps to the current (not in this PR) ServerSettings, which are a list of keys/values. One thing that I did with this PR is separated the config into two distinct categories - config for the Immich account(s) that you are accessing, versus config for ImmichFrame. This was necessary when adding multiple accounts.

The current code does the best it can when loading from a flat list (environment variables), by prefixing them with Account2. etc. However JSON does not need to be a flat list. We need to make a choice on a few things:

- Do we expose the split in account versus server config in JSON? Or with env vars?
- Do we nest the distinct account settings in JSON (group them into their own JSON object)?

Note that if we do not choose yes to both of these, then we need custom JSON loading as our (now nested) config structure will no longer match the JSON. This is fine, I have no problem implementing this, but wanted to let you know in case you want to avoid divergence of config class structure and config file structure.

jpg0:
- PR to added total asset count to immich search: https://github.com/immich-app/immich/pull/18885
- PR to allow querying by album https://github.com/immich-app/immich/pull/18552 (this should also be enabled for asset counts)
-  I will open another PR for asset counts for memories, once I get feedback on the first one

JW-CH:
> I have rebased this onto the fixed config loading, and addressed the points above. However, I have not yet implemented multiple accounts via JSON config. I can do so, but I need more direction on where you want the config structure to go:
> 
> Currently the structure is flat - a simple list of keys. This maps to the current (not in this PR) ServerSettings, which are a list of keys/values. One thing that I did with this PR is separated the config into two distinct categories - config for the Immich account(s) that you are accessing, versus config for ImmichFrame. This was necessary when adding multiple accounts.
> 
> The current code does the best it can when loading from a flat list (environment variables), by prefixing them with Account2. etc. However JSON does not need to be a flat list. We need to make a choice on a few things:
> 
> * Do we expose the split in account versus server config in JSON? Or with env vars?
> * Do we nest the distinct account settings in JSON (group them into their own JSON object)?
> 
> Note that if we do not choose yes to both of these, then we need custom JSON loading as our (now nested) config structure will no longer match the JSON. This is fine, I have no problem implementing this, but wanted to let you know in case you want to avoid divergence of config class structure and config file structure.

Good question. I am not quite sure how to handle this the best way. My first thought was to just have the 'old' format with only one account be able to deserialized and put in the new format and support for the 'new' format. Meaning we try to parse the JSON in the new format, if that fails we try with the old one and map it to the new one. Not sure if this is ideal, but this would be backward compatible. I know that some people use the JSON config, so this is not something we should just get rid of. @jpg0 you seem quite experienced, what's your take on this?

3rob3:
I’m ok dropping the json support.

jpg0:
Regarding config formats, I would suggest that internally we use the most powerful structure, and adapt any older or less capable formats to it. So this would mean that we keep the class format in this PR.

I would also suggest that in the interest of keeping existing installations working, we retain the ability to load both the flat JSON and environment variables for at least one major revision. As mentioned these should be with adapters that 'bolt-on' the functionality without holding back the main code. Then we'd need to make a decision on whether we deprecate either of these formats, and potentially remove the adapters later.

I admit that whilst loading many properties from env vars is nice as it avoids a config file, it's uncommon to see this much config loaded this way. I believe that it's because it's ultimately quite limiting in what it can do. I'd also point out that having two mechanisms to define config is also uncommon (but not unheard of). 

If it was me I'd probably:
- Make structured JSON the preferred config file
- Keep flat JSON and env vars working via adapters
- Deprecate their usage in the docs, but keep them working via the adapters
- Only remove them later if and only if they get in the way of another feature

Happy to implement this (well, the first 3 points) if you'd like me to.

3rob3:
Agree that would be ideal, but I’d also argue this isn’t an app that would seriously hose people if it broke. Worse case they roll back until they’re ready to switch over (with a big breaking change release note of course).  I vote powerful and simplicity code wise. 

Also, thank you for your contributions. You really seem knowledgeable in this.

jpg0:
Maybe I've spent too much time on commercial codebases where you can't break things! Of course I'd also be happy to to just make this a breaking change as it's less work for me :)

jpg0:
The immich PR is approved, we just need it to be released. I will also open another PR to add totals for memories too, at which point I think we'll have covered all the functionality here so can correctly balance between accounts, plus avoid pre-caching any assets within any account. Of course this will raise an additional question: do we support older versions of immich too, or do you just state that it's another breaking change and you need the latest version of immich? Maybe we should just make a clean break in both config and supported immich versions and called this 2.0?

3rob3:
> The immich PR is approved, we just need it to be released. I will also open another PR to add totals for memories too, at which point I think we'll have covered all the functionality here so can correctly balance between accounts, plus avoid pre-caching any assets within any account. Of course this will raise an additional question: do we support older versions of immich too, or do you just state that it's another breaking change and you need the latest version of immich? Maybe we should just make a clean break in both config and supported immich versions and called this 2.0?

I'd again vote marking it as breaking change/latest immich version required. Might depend on how long it takes for them to release though?

JW-CH:
I'd also say it is fine to only support the newest version of immich with a release on our side. Not sure about the `2.0`, that's maybe something for the immich stable release? Regarding the config stuff: I'd say we should try to stay backward compatible with the config. Having to redo your config could be upsetting.

jpg0:
Asset counts PR is now merged in immich, and I've just opened another one (https://github.com/immich-app/immich/pull/19035) to give memory counts. These should be sufficient to avoid pre-loading anything but these totals.

I'm happy to add backwards compatibility for the config to this PR, I'll keep it as separate as possible so that @3rob3 can kill it soon!

Maybe then we can then merge (multi-account with them work but only based on total # of assets, certainly not worse than we have now). I will then open another PR which targets an updated immich, and either we wait until an updated immich is released, or I include backwards compatibility there too.

jpg0:
I've started on the config stuff, with fallbacks to the original style. One question I have is about `IWebClientSettings`: these seem like they overlap but are not the same as other settings. Initially I just thought that I could get rid of them and merge into the main settings (there are only 3 properties that are actually used), however I see that they are exposed by `GetConfig` in the API. Does this need to be kept the same? If not I can update to the nested config, otherwise I could still merge the settings' types but adapt the new settings to this original style?

(Normally the pattern here would be to create a standard settings class, then adapters to create restricted or modified 'views' into this main one.)

Also I am adding unit tests to ensure that original config files continue to keep working.

3rob3:
The IClientSettings are passed to/used in the non-webview clients (Apple TV and optionally in Android). Hopefully these don't need to be changed, or at least the API returns the values in the same format.

jpg0:
I can certainly keep them returning the same data. However, why is `ImmichServerUrl` in there? I can see that it's not used by the Android client, but I don't know where the Apple TV source is (that one is awkward as there can now be multiple Urls).

3rob3:
Apple TV uses same as Android non webview. Client wise I am not using immichserverurl, not sure why that would be in there. I can give you access to Apple TV if you want/need. It could be public but I’m new at that and was afraid of exposing sensitive data. I made that mistake with Android at the beginning 😬.

jpg0:
No, I don't need access, it would only to see if the `ImmichServerUrl` is used. As it's not I'll just remove it from the API response.

3rob3:
> No, I don't need access, it would only to see if the `ImmichServerUrl` is used. As it's not I'll just remove it from the API response.

I don't mind :).  
This are the settings I use on Apple TV:

struct ServerSettings: Codable {
        let margin: String?
        let interval: Int
        let transitionDuration: Double
        let downloadImages: Bool
        let renewImagesDuration: Int
        let showClock: Bool
        let clockFormat: String?
        let showPhotoDate: Bool
        let photoDateFormat: String?
        let showImageDesc: Bool
        let showPeopleDesc: Bool
        let showImageLocation: Bool
        let imageLocationFormat: String?
        let primaryColor: String?
        let secondaryColor: String?
        let style: String
        let baseFontSize: String?
        let showWeatherDescription: Bool
        let imageZoom: Bool
        let imageFill: Bool?
        let layout: String
        let language: String
    }
    
    Pretty much same on Android non webview (unattended is there but unused):
    
    data class ServerSettings(
        val margin: String,
        val interval: Int,
        val transitionDuration: Double,
        val downloadImages: Boolean,
        val renewImagesDuration: Int,
        val showClock: Boolean,
        val clockFormat: String,
        val showPhotoDate: Boolean,
        val photoDateFormat: String,
        val showImageDesc: Boolean,
        val showPeopleDesc: Boolean,
        val showImageLocation: Boolean,
        val imageLocationFormat: String,
        val primaryColor: String?,
        val secondaryColor: String,
        val style: String,
        val baseFontSize: String?,
        val showWeatherDescription: Boolean,
        val unattendedMode: Boolean,
        val imageZoom: Boolean,
        val imageFill: Boolean,
        val layout: String,
        val language: String
    )

jpg0:
Ok, I have added backwards compatibility to the config (original JSON + env vars), plus the API. Settings have been reworked a little (there are accounts + general settings, all in a server settings object). I have updated the docs to use the new JSON format, and deprecated the old format. I've removed the docker instructions for the old format, and added a sample JSON file for the new format. I've squashed the commits into one, and rebased from main.

Please take a look, there's a lot in this branch now so I'm keen to get it merged so that it doesn't diverge further!

Also, the final immich PR was merged too, so once this is merged I'll begin consuming the new immich code to prevent the asset preloading.

dohlin:
Soooo excited to see this feature coming, been wanting to combine mine and my wife's photos on a single digital picture frame for ages! Can't wait!!

JW-CH (approved):

JW-CH:
Thank you very much for your work @jpg0, it is much appreciated ❤️🚀

jpg0:
No worries! There was quite a few changes in there, so let me know if you do find a regression anywhere. I did add some tests, but they are not yet being run by github (I'll add a PR for that).

Next I'd like to move towards the "no preload" strategy now that we have the changes in immich:
- Modify the existing code to operate with multiple asset 'pools' (e.g. memories may be a pool), where the pool can provide it's total # of assets, or a selection of random ones - initially this will be via a preload).
- Have the main logic aggregate assets across pools, probabilistically based on the number in each pool
- Simplify multi-account support by just using account-specific pools
- Swap out pool implementations to 'no-preload' ones where possible (probably retain a fallback to preload too)

Hope that makes sense!

dohlin:
Stupid question, but just to be sure - the updated settings.json with multiple accounts will not be usable until an updated immich-frame docker has been published, correct?

JW-CH:
> Stupid question, but just to be sure - the updated settings.json with multiple accounts will not be usable until an updated immich-frame docker has been published, correct?

Correct, we will create a new release in the near future :) You'll need to wait for that.

JW-CH:
@jpg0 Do you remember why the cache is set to 60 seconds? #408

https://github.com/immichFrame/ImmichFrame/blob/7ddc90aa2c8db3e447d1ffc81204de8c230246d2/ImmichFrame.Core/Logic/MultiImmichFrameLogicDelegate.cs#L19

jpg0:
No specific reason for 60 sec. It caches all the assets it hands out (to be able to map back to accounts). I assumed that if a client was going to ask again for data on an asset then it would either be immediately (in which case 60 sec should be fine), or potentially much longer (in which case any number may not be). Happy to change it though, and/or add a backup option to iterate through accounts if it's not found. (The other option is to use composite keys or guid+account but that's more work and would affect the clients.)

jpg0:
Was just thinking about this and had another idea: I could use bloom filters to track where the guids came from. This could probably strike a much better balance between memory usage and performance (assuming falling back to iterating accounts).

jpg0:
Given this, I decided to replace the existing cache with bloom filters - I've added it to the asset-pools PR, it should perform much better!

JW-CH:
@jpg0 hi, sorry me again. I just noticed that the `/api/config` endpoint outputs the client secret due to the changes made to the settings. This was the reason I had `IClientSettings` and `IServerSettings`. I am currently thinking of a way to fix this. The Secret should never be output!

jpg0:
> @jpg0 hi, sorry me again. I just noticed that the `/api/config` endpoint outputs the client secret due to the changes made to the settings. This was the reason I had `IClientSettings` and `IServerSettings`. I am currently thinking of a way to fix this. The Secret should never be output!

I like your original approach - I just didn't understand the difference so assumed that it was unused. Happy to go back to that - just with either docs or name the interface as IExposedSettings or something so that it's obvious why it's done that way? I'm happy to implement if you like?

3rob3:
IExposedSettings sounds good to me. I think he may have started working on this change already, but not sure about that.

jpg0:
> I think he may have started working on this change already, but not sure about that.

Ok, I'll wait.

JW-CH:
> > I think he may have started working on this change already, but not sure about that.
> 
> Ok, I'll wait.

I threw that away. Did not have time yesterday and probably won't have today.

jpg0:
Ok, I broke it so I'll fix it!

jpg0:
https://github.com/immichFrame/ImmichFrame/pull/411

PR #376, opened 2025-05-31
Chore: Cleanup fire-and-forget asset loading

jpg0:
OptimizedImmichFrameLogic has an optimisation which loads a queue of assets that can be pulled from to 'GetNextAsset'. There were a few of problems in the code:

* The initial call would always return null, because the queue would be empty -> now it async reloads just as before, but also waits (up to 1min) to the fetch to happen to return an asset (and only returns null if it takes < 1min)
* The code used a standard Queue which is not thread-safe (so could have failed with concurrent modifications) -> replaced with a thread-safe Channel
* The code attempted to prevent concurrent fetches (wait for an existing to complete rather than start a 2nd), but guarded via reading an instance-level bool (where writes may not be observed immediately by another thread) -> replaced with a SlimSemaphore


JW-CH (approved):
Really clean, I am learning with your commits 😄

PR #377, opened 2025-06-01
Refactor: Use interface implementations in DI

jpg0:
Switch to implicit DI instance creation

### DIFF
--- a/ImmichFrame.WebApi/Program.cs
+++ b/ImmichFrame.WebApi/Program.cs
@@ -65,43 +65,11 @@ builder.Services.AddLogging(builder =>
     builder.AddFilter("Microsoft.AspNetCore", LogLevel.Warning);
 });
 
-builder.Services.AddSingleton<IServerSettings>(srv =>
-{
-    if (serverSettings == null)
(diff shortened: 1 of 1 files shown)


JW-CH (approved):
Keeping it clean 👌🏻

ImmichFrame.WebApi/Program.cs
```diff
@@ -65,43 +65,11 @@ _ _ __ ___  _ __ ___  _  ___| |__ | |_ _ __ __ _ _ __ ___   ___
     builder.AddFilter("Microsoft.AspNetCore", LogLevel.Warning);
 });
 
-builder.Services.AddSingleton<IServerSettings>(srv =>
-{
-    if (serverSettings == null)
-        serverSettings = new ServerSettings();
-
-    return serverSettings;
```
JW-CH, line 73: @jpg0 Actually now that I am thinking about this. This breaks the current functionality of having a json or yaml file as settings.

jpg0: Good catch! I was not aware that this is supported, but see it now in the docs - I'll fix this.

PR #384, opened 2025-06-03
Update example.env with comments and reorder

arcoast:
Recently setup Immich Frame and it's awesome, I found it useful to reorder the variables and put into categories in the `example.env` and comment it to reflect the current docs to improve usability when setting up.

Not sure if you'll be interested in merging this and feel free to close if not, but figured as I'd done it for myself I'd submit it for consideration.

Thanks for the project, it's awesome!

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -1,40 +1,150 @@
+# See docs here https://immichframe.online/docs/getting-started/configuration
+
+# Required
+## The URL of your Immich server e.g. http://photos.yourdomain.com / http://192.168.0.100:2283.
+
 ImmichServerUrl=URL
+
+## Read more about how to obtain an Immich API key here https://immich.app/docs/features/command-line-interface/#obtain-the-api-key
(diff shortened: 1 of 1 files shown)


JW-CH:
Thank you for your contribution. I like the comments, but my fear is that those will be outdated as the project continues to grow. With every duplicate of those settings/comments, they will get out of date or forgotten to change, that is why I kept them as minimal as possible. What's your take on this @3rob3 

Maybe we could automate this?

arcoast:
I respect that, like I said, I'd made the changes for myself, so thought I'd offer it up, but I'm not going to be upset or offended if you want to keep it simple, whatever you feel is the best for the project I'm totally fine with!

:+1:

3rob3:
Agree with everything above. I think the proposed changes look really nice, but we kinda had it more like this before and found simplifying them was easier to maintain across the different settings options.
I do appreciate the contribution and think they look better.

Issue #135, opened 2024-09-11
Animate photo size/location


chrisscotland:
### Description

Where photos don't fit the screen aspect ratio, or where it is otherwise desired they could be animated so they scroll across the screen.

### Use Case

Photo is landscape and screen is portrait. Current uniformtofill display mode results in only the middle of the image being displayed. If it was animated it could slowly scroll across the image for the duration the image is on screen.

### Details

A smooth animated feature would be useful for a variety of image sizes. Scroll could be horizontal or vertical and automatically selected depending on current image and screen aspect ratios. A zoom mode could also work where images zoom in or out slowly and this could apply even if screen and photo aspect ratios match. It would make the images appear more dynamic and the overall experience a bit more engaging.

### Pre-Submission Checklist

- [X] This is a feature request and not a bug report

- [X] I have checked for related issues and checked the documentation

3rob3:
This is more complex than you would imagine, at least with the framework we are using. I think it would be a cool option though.

3rob3:
We added a zoom effect option in latest version. If you feel this satisfies your issue can you close this?

chrisscotland:
Thanks very much the new effect works really well and I feel the result looks very good 😊

My original suggestion was really regarding images where the aspect ratio doesn't match with the screen size. That could still be improved with a scrolling animation (either horizontally or vertically over the image). I appreciate this might be harder to implement though.

jinwater88:
@3rob3 What technology is used to make this? Moviepy?

JW-CH:
> [@3rob3](https://github.com/3rob3) What technology is used to make this? Moviepy?

@jinwater88 The web version is using CSS to pan/zoom.

PR #391, opened 2025-06-05
Feat: Pan option

pete1450:
Adds a subtle random pan direction if enabled with `ImagePan: "true"`

Defaulted to false for now. It does allow for zoom and pan as well.


JW-CH:
Closes #135

JW-CH:
Thank you for your PR, could you also update the docs (add option to the table) and add it to the config examples:

https://github.com/immichFrame/ImmichFrame/blob/main/docs/docs/getting-started/configuration.md
https://github.com/immichFrame/ImmichFrame/blob/main/docs/docs/getting-started/installation/docker.md
https://github.com/immichframe/ImmichFrame/blob/main/docker/Settings.example.json
https://github.com/immichframe/ImmichFrame/blob/main/docker/example.env

JW-CH:
image
The pan offsets the image, which I don't like too much. Maybe it would be possible to have a default zoom on a pan image or as described in #135, only pan when an image is too large.

pete1450:
Heh I noticed the offset and wondered what other opinions would be. I was going for an animation all the time so I'll play with a default zoom. Docs on the list as well.

pete1450:
@JW-CH To double check we're seeing the same thing; you only see the offset when you have both zoom and pan enabled correct?

JW-CH:
> @JW-CH To double check we're seeing the same thing; you only see the offset when you have both zoom and pan enabled correct?

Nope, the offset comes from the pan setting and is also visible with the pan only enabled. My guess would be the `-5%` from the `--pan-start` and the `--pan-end`, but not 100% sure.

PR #402, opened 2025-06-14
Feat: Support yaml config & Update Docs

JW-CH:
I hate that JSON has no support for comments - YAML does.

TODO: Add example & Update Docs


JW-CH:
@jpg0 I'd be interested in your review if you find some time.

3rob3:
Somebody just recently asked for this, I can’t find where now. Nice!

ImmichFrame.WebApi/Helpers/Config/ConfigLoader.cs
```diff
@@ -3,30 +3,59 @@
...
+            }
+
+            try
+            {
+                var v1 = LoadConfigYaml<ServerSettingsV1>(ymlConfigPath);
+                return new ServerSettingsV1Adapter(v1);
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning("Failed to load config as old JSON, falling back to env vars (${errorMessage})", e.Message);
```
jpg0: Failed to load config as old YAML?

ImmichFrame.WebApi/Helpers/Config/ConfigLoader.cs
```diff
@@ -3,30 +3,59 @@
...
-            var v1 = LoadConfigJson<ServerSettingsV1>(filename);
-            return new ServerSettingsV1Adapter(v1);
+            try
+            {
+                var v1 = LoadConfigJson<ServerSettingsV1>(jsonConfigPath);
+                return new ServerSettingsV1Adapter(v1);
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning("Failed to load config as old JSON, falling back to env vars (${errorMessage})", e.Message);
```
jpg0: Failed to load config as old JSON, trying as YAML?

jpg0 (approved):
Only high level thing that I would suggest is whether you want to support the flat structure with YAML, or just go straight to the structured one. If we do remove the flat one at some point, you don't want people having to migrate YAML when they could have gone straight to the V2 version.

Also don't forget docs!

JW-CH:
> Only high level thing that I would suggest is whether you want to support the flat structure with YAML, or just go straight to the structured one. If we do remove the flat one at some point, you don't want people having to migrate YAML when they could have gone straight to the V2 version.
> 
> Also don't forget docs!

Good point with the v1, does not make sense for people to use that. Docs are on the to-do. Thank you for the review.

ImmichFrame.WebApi/Helpers/Config/ConfigLoader.cs
JW-CH: Hmm, I added a check if the config file exists. I'd have to check if the yml exists in the json check, but this looks ugly to me.

3rob3: I think he meant the catch _logger message. Should be YAML not "env vars"

ImmichFrame.WebApi/Helpers/Config/ConfigLoader.cs
```diff
@@ -3,30 +3,58 @@
...
         {
-            _logger.LogWarning("Failed to load config as current version, falling back to old version (${errorMessage})", e.Message);
-        }
+            try
+            {
+                return LoadConfigJson<ServerSettings>(jsonConfigPath);
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning("Failed to load config as current version, falling back to old version (${errorMessage})", e.Message);
```
3rob3: I might be wrong, but shouldn't this be "{errorMessage}", not "${errorMessage}"? Same with others below.

docs/docs/getting-started/configuration.md
```diff
@@ -138,9 +135,9 @@ A client can be identified by the `ClientIdentifier`. You can set/overwrite the
...
-| -------------------------- | --------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| ImageRequestedNotification | Notifies, when a Image requested. | `{"Name":"ImageRequestedNotification", "ClientIdentifier": "Frame_Kitchen", "DateTime":"2024-11-16T21:37:19.4933981+01:00", "RequestedImageId":"UUID"}` |
+| **Event**                  | **Description**                      | **Payload**                                                                                                                                             |
+| -------------------------- | ------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| ImageRequestedNotification | Notifies when an image is requested. | `{"Name":"ImageRequestedNotification", "ClientIdentifier": "Frame_Kitchen", "DateTime":"2024-11-16T21:37:19.4933981+01:00", "RequestedImageId":"UUID"}` |
 
 ### Multiple Immich Accounts
```
3rob3, line 141: I think we need more detail on using this, maybe just a "see settings.example.x for more details?

docs/docs/getting-started/configuration.md
```diff
@@ -7,111 +7,111 @@ sidebar_position: 2
 ### Full configuration reference:
 
 :::warning
-It is not recommended to copy this full configuration file. Only specify values that are different from the defaults. Configuration options and default values may change in future versions.
+It is not recommended to copy this full configuration file.
+Only override settings you intend to change.
+Defaults might change between versions, so keeping your config minimal helps future upgrades.
 :::
```
3rob3, line 14: Should we have links to settings.example.json and yml here instead of the previous page?

ImmichFrame.WebApi.Tests/Resources/TestV1.json
```diff
@@ -53,6 +52,5 @@
     "Account2.ImmichServerUrl": "Account2.ImmichServerUrl_TEST",
     "Account2.ApiKey": "Account2.ApiKey_TEST",
     "Account2.ImagesFromDate": "Account2.ImagesFromDate_TEST",
-    "Margin": "Margin_TEST",
     "UnattendedMode": true
```
3rob3: Not directly related to this PR, but since you were already removing unused settings :), I think UnattendedMode can also be removed.

ImmichFrame.WebApi/Helpers/Config/ConfigLoader.cs
jpg0: Yes, I was referring the message (looked copy-pasted, but the details change as you progress through the file types).

ImmichFrame.WebApi/Helpers/Config/ConfigLoader.cs
jpg0: You're right, the whole log formatting versus string interpolation is frustrating similar.

PR #404, opened 2025-06-15
refactor: break out asset sources into distinct composable pools

jpg0:
Refactored asset sourcing into distinct, composable pools. This should not affect current functionality, but allows us to now add non-preloading pools easily and safely. (I guess it would also allow pulling from non-immich sources too, but that's not the goal.)

There are quite a few tests included. These were mostly generated by Google Jules, but I have gone over them to make sure that they are testing the right things etc. The tests could certainly be improved (making more DRY for example), but they work well and it may be better to wait for the pool changes to be complete before optimising them.


jpg0:
FYI I have also begun a branch based on this change to remove preloading where possible. So far for the various asset pools:
- When unfiltered, there is no preloading (just as before)
- Favorites can easily be moved to no preloading (I have this already in a branch)
- Memories cannot yet be moved, as there is no mechanism to randomly search for them (I can add this to Immich though)
- Albums and People cannot yet be moved, because by default immich searches AND the terms together (intersection of results), whereas we OR them (union of results). So there is no way to search in immich for photos which contain person 1 OR person 2. However there is an PR for this: https://github.com/immich-app/immich/pull/18552, so assuming that it gets merged, it will unlock these pools too (or at least albums, but people is kinda the same so will either be included, followed, or I can add it)

Note that with this refactor the fetching is done by each pool, and each pool can operate independently, so we can add support to each pool as it becomes possible.

JW-CH:
Thank you for the implementation and your dedication for new features/refactorings to this project. I will try my best to review this as soon as possible.

jpg0:
FYI, just also added improved asset tracking across accounts using bloom filters.

MultiAssetPoolTests.cs
```diff
@@ -0,0 +1,526 @@
...
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+// Assuming the classes are in a namespace, adjust if necessary
+// namespace YourNamespace.Tests;
+
+[TestFixture]
+public class MultiAssetPoolTests
```
JW-CH: Is this some leftover file? This looks like it does not belong here.

AggregatingAssetPoolTests.cs
```diff
@@ -0,0 +1,248 @@
+using NUnit.Framework;
+using Moq;
+using System.Collections.Generic;
+using System.Linq;
+
+// Assuming the classes are in a namespace, adjust if necessary
+// namespace YourNamespace.Tests;
+
+[TestFixture]
+public class AggregatingAssetPoolTests
```
JW-CH: Is this some leftover file? This looks like it does not belong here.

JW-CH:
Also: Run tests with GitHub CI, maybe with a filter? Only run when Core/WebApi has changed.

MultiAssetPoolTests.cs
jpg0: weird, you're right, I guess Jules left it there, I'll remove it

jpg0:
> Also: Run tests with GitHub CI, maybe with a filter? Only run when Core/WebApi has changed.

I already had the WebApi tests running, I've now added Core too.

jpg0:
> I already had the WebApi tests running, I've now added Core too

And... they are failing. Good call to add them to CI! I'll check out the failures.

JW-CH:
I love the changes! Genius using bloom filters (had to look those up first, lol). TBH, I did not 100% understand the tests as of now (Mocking etc.), but the code looks really well. I assume that with a future release of Immich, we will be able to switch the pools from a `CachingApiAssetsPool` to a `QueuingAssetPool`, correct?

JW-CH (approved):

JW-CH:
Closes #408

jpg0:
> Genius using bloom filters (had to look those up first, lol). 

I have known about them for a long time, but this is the first time I've actually found a use for them!

> TBH, I did not 100% understand the tests as of now (Mocking etc.), but the code looks really well. 

Hope that my code helps your learning! The primary purpose of using a DI framework is so that components do not directly depend on each other, so that the app is "assembled" by the DI framework, externally to the components themselves, which then enables testing each component independently. When doing this, components often need some implementation of their dependencies to operate at all. There are various [test doubles](https://martinfowler.com/bliki/TestDouble.html) that can be used, in this case we are using mocks which allow us to assert that dependencies provide the right data to the class being tested, and that the classes call the right things on their dependencies.

> I assume that with a future release of Immich, we will be able to switch the pools from a CachingApiAssetsPool to a QueuingAssetPool, correct?

Even better - we can create a 'FallbackAssetsPool' or similar that delegates to either of these pools - if the queueing works (immich supports the API being called) then it will use this, otherwise it will fall back to a caching api pool. This means that we will always use the optimal immich API where possible, but it will still work if only the old API is available. At some point in the future (when almost noone is on a version of immich without the new APIs) we remove all the caching apis entirely.

ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public Task<IEnumerable<AssetResponseDto>> GetAssets()
+    {
+        return _pool.GetAssets(25);
+    }
+
+    public Task<AssetResponseDto> GetAssetInfoById(Guid assetId) => _immichApi.GetAssetInfoAsync(assetId, null);
+
+    public async Task<IEnumerable<AlbumResponseDto>> GetAlbumInfoById(Guid assetId) => await _apiCache.GetOrAddAsync("GetAlbumInfoById",
```
JW-CH: This cache should not be here, right?

JW-CH (changes requested):

ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
jpg0: Ah yes. I admit that this interface (IImmichFrameLogic) does mix both pool handling plus looking up additional data from guids; I missed this one. Just pushed a fix.

jpg0:
New immich is now released with the asset counts code :)

PR #410, opened 2025-06-18
Docs: add general instructions for updating WebView

proud-nerd:
Here is a little documentation how to update WebView on devices other than Android 6.01 10.1" Frameo, as you asked for in #406

### DIFF
--- a/docs/docs/getting-started/apps.md
+++ b/docs/docs/getting-started/apps.md
@@ -120,7 +120,7 @@ If this worked for you, consider opening a 'Show and tell'-Discussion.
 
 ```
 ❗️ Warning:
-This might break your warranty
+This might break your warranty.
 ```
 
 The basic procedure of updating the WebView installation consists of a few steps. Read through and understand them first, before you attempt any changes.
--- a/docs/docs/getting-started/apps.md
+++ b/docs/docs/getting-started/apps.md
@@ -108,10 +108,73 @@ ADB is often enabled on these devices by default, if it is not go to Frameo Sett
       adb uninstall com.immichframe.immichframe
       ```
   
-### Frameo WebView Update  
-Follow instructions below to update WebView to 106. This has been tested and working on Android 6.01 10.1" Frameo devices:  
+### Frameo WebView Update
+
+The basic procedure of updating the WebView installation consists of a few steps. Read through and understand them first, before you attempt any changes.
+
+To make it easier to follow, there are tested and working examples for Android 6.01 10.1" Frameo devices provided at the end. If you have one of those, you might as well skip directly to them.
+
+- Download WebView to PC from [APKMirror (WebView by Lineage)][webview-lineage]
+  - You most likely must use an APK with com.android.webview as the package name, that's why LineageOS version is recommended instead of the Google version (which is com.google.android.webview)
+  - You must use a version that is compatible with your Android version
+  - You must use a version that is compatible with your device's architecture
+- Find the current WebView APK on your device
+  ```shell
+  adb shell pm path com.android.webview
+  ```
+  might return something like `/system/app/webview/webview.apk` or `/product/app/webview/webview.apk`
+- Rename downloaded APK to match the original one (compare with output of the command above)
+  ```
+  mv downloaded-webview.apk webview.apk
+  ```
+- Push new apk to sdcard
+  ```shell
+  adb push /path/to/your/new/webview.apk /sdcard/
+  ```
+- Enter root shell
+  ```shell
+  adb shell su
+  ```
+- Mount the correct partition as read-write. Replace `/partition` with the first path of the original WebView APK on your device
+  ```shell
+  adb shell mount -o rw,remount /partition
+  ```
+- Backup original WebView APK
+  ```shell
+  mount -o rw,remount /partition && cp /path/to/original/webview.apk /path/to/original/webview.apk.bak
+  ```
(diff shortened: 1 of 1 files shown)


docs/docs/getting-started/apps.md
```diff
@@ -108,10 +108,73 @@ ADB is often enabled on these devices by default, if it is not go to Frameo Sett
...
+  ```
+- Mount the correct partition as read-write. Replace `/partition` with the first path of the original WebView APK on your device
+  ```shell
+  adb shell mount -o rw,remount /partition
+  ```
+- Backup original WebView APK
+  ```shell
+  mount -o rw,remount /partition && cp /path/to/original/webview.apk /path/to/original/webview.apk.bak
+  ```
+- Override original WebView APK with the downloaded one
```
3rob3: Overwrite

docs/docs/getting-started/apps.md
```diff
@@ -108,10 +108,73 @@ ADB is often enabled on these devices by default, if it is not go to Frameo Sett
...
+  ``` 
+- Delete the oat folder recursively 
+  - needed for Android 6.1 and most likely for all devices, where WebView is installed as a system app below `/system`
+  - might not be needed/possible if original WebView is installed for example below `/product`
+  ```shell
+  mount -o rw,remount /system && rm -rf /system/app/webview/oat
+  ```
+- Copy new WebView to system
+  ```shell
+  mount -o rw,remount /partition && cp /sdcard/webview.apk /path/to/original/webview.apk
```
3rob3: Why are we doing this twice?

docs/docs/getting-started/apps.md
```diff
@@ -108,10 +108,73 @@ ADB is often enabled on these devices by default, if it is not go to Frameo Sett
       adb uninstall com.immichframe.immichframe
       ```
   
-### Frameo WebView Update  
-Follow instructions below to update WebView to 106. This has been tested and working on Android 6.01 10.1" Frameo devices:  
+### Frameo WebView Update
```
3rob3, line 111: Should this header indicate Android 7+ or something?

docs/docs/getting-started/apps.md
```diff
@@ -108,10 +108,73 @@ ADB is often enabled on these devices by default, if it is not go to Frameo Sett
...
+
+The basic procedure of updating the WebView installation consists of a few steps. Read through and understand them first, before you attempt any changes.
+
+To make it easier to follow, there are tested and working examples for Android 6.01 10.1" Frameo devices provided at the end. If you have one of those, you might as well skip directly to them.
```
3rob3: this is kind of confusing. Do you mean the older instruction for Android 6?

3rob3:
Thx for contributing! Just a couple changes requested.

docs/docs/getting-started/apps.md
proud-nerd: whoops, missed that one. I did quite some restructuring iterations :laughing: good catch :+1:

docs/docs/getting-started/apps.md
proud-nerd: Yeah, my idea was to have this section for general instructions how to do it on any device/android version and still leave the old - but tested and working - instructions as a subsection at the end.

There is quite some "path/to/x/apk" in the general instructions and the whole mounting the correct partition thing, so it might be a little confusing for some. That's why I thought it might be helpful to have those "older" but very specific instructions with actual paths as a reference.

But I can see how it's confusing if we say "read and understand those first" and then "you might skip directly to the commands" :laughing: So I'd propose I just rewrite that part, but I'd still want to mention the instructions for those specific devices.

Would that work for you or do you have any more suggestions how to make this clearer?

proud-nerd:
Thanks for your feedback. I fixed some of your catches, but I need some more feedback on that confusing part. I'll provide a new proposal after that, that incorporates your feedback.

proud-nerd:
I'd also want to add a disclaimer, that this might not work for every device, as there might be some vendor lockdowns in place. Also people should be aware, that they might be losing warranty with this procedure, even though I'm not sure about that, and should be ready to restore their WebView from backup or reset their device in case of failure.

In the end, this _should_ work for all possible kinds of devices, but as there are a myriad of different android versions and devices out there, people should just be aware that it might not be smooth sailing or not even work at all.

Would you approve such a disclaimer?

3rob3:
> I'd also want to add a disclaimer, that this might not work for every device, as there might be some vendor lockdowns in place. Also people should be aware, that they might be losing warranty with this procedure, even though I'm not sure about that, and should be ready to restore their WebView from backup or reset their device in case of failure.
> 
> In the end, this _should_ work for all possible kinds of devices, but as there are a myriad of different android versions and devices out there, people should just be aware that it might not be smooth sailing or not even work at all.
> 
> Would you approve such a disclaimer?

Sure. Just try to keep everything as brief as possible.

proud-nerd:
I added the disclaimer and changed some bits to (hopefully) make things clearer.

proud-nerd:
Anything else I can/should do, to bring this to an approved state?

3rob3:
I like the disclaimer and warning, and think we should add the "Find the current WebView APK on your device" step (and to replace that path in the other commands), but I feel like the rest is too wordy and repeated without adding much. What are your thoughts on just doing those minimal updates?

proud-nerd:
Sure, I can do that. Sounds good to me, it is quite repetitive right now.

PR #411, opened 2025-06-19
Restrict authenticationSecret from being serialised in API

jpg0:
Added a custom type resolver to omit the AuthenticationSecret field when an IGeneralSettings is being serialised by ASP.NET.

This works for a single omission like this, but I would recommend creating DTO types if there is any more divergence.

### DIFF
--- a/ImmichFrame.Core/Interfaces/IServerSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IServerSettings.cs
@@ -1,4 +1,9 @@
-﻿namespace ImmichFrame.Core.Interfaces
+﻿using System.Reflection;
+using System.Text.Json;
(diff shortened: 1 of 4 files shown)


JW-CH:
The client side API gets generated via the `OpenAPI`-Specification.

1. The OpenApi still generates the field `AuthenticationSecret`
2. More fields are sent via the config-endpoint that should not appear there, see image. (WeatherApiKey, Webhook, Calendars)

see: 
image

jpg0:
Ok, in this case I'll switch to a DTO. I'll use the same fields as present in the original `IClientSettings`.

jpg0:
Superceded by https://github.com/immichFrame/ImmichFrame/pull/412

PR #412, opened 2025-06-19
Added ClientSettingsDto to expose subset of general settings in API

jpg0:
Implemented ClientSettingsDto to provide subset of general settings to clients. Using a DTO means that further evolution of internal model won't change API. Note that `ImmichServerUrl` is omitted as there may now be multiple (and we previously confirmed that it was not used).


3rob3:
LGTM

PR #416, opened 2025-06-26
Fix: Open in immich

JW-CH:
This PR fixes the 'open in immich' button that was broken since the multi source for immichframe. This PR adds a `ImmichSourceUrl`-Property to the `AssetResponseDto`. The change allows the webclient to open the correct url.

cc @jpg0

### DIFF
--- a/ImmichFrame.Core/Api/AssetResponseDto.cs
+++ b/ImmichFrame.Core/Api/AssetResponseDto.cs
@@ -9,6 +9,8 @@ namespace ImmichFrame.Core.Api
         [JsonIgnore]
         public Stream? ThumbhashImage => GetThumbHashStream();
 
+        public string ImmichSourceUrl { get; set; } = string.Empty;
+
         private Stream? GetThumbHashStream()
         {
             if (this.Thumbhash == null)
--- a/ImmichFrame.Core/Logic/Pool/AlbumAssetsPool.cs
+++ b/ImmichFrame.Core/Logic/Pool/AlbumAssetsPool.cs
@@ -24,6 +24,8 @@ public class AlbumAssetsPool(ApiCache apiCache, ImmichApi immichApi, IAccountSet
             albumAssets.AddRange(albumInfo.Assets);
         }
 
+        albumAssets.ForEach(t => t.ImmichSourceUrl = accountSettings.ImmichServerUrl);
+
         return albumAssets.WhereExcludes(excludedAlbumAssets, t => t.Id);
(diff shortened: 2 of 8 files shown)


jpg0:
Apologies - I did ask about it ImmichServerUrl in https://github.com/immichFrame/ImmichFrame/pull/374#issuecomment-2960945554 and subsequently removed it.

jpg0:
Regarding the actual fix here, I'd suggest that the `ImmichServerUrl` is added to the dtos by the code which selects between the accounts to draw the assets from (in this case `TotalAccountImagesSelectionStrategy`).  This way additional logic pools etc can be added or updated without needing to care about this, and the code to add the `ImmichServerUrl` (which comes from the config rather than the asset itself) is only in a single place, which is already a class that is concerned with the links between accounts and assets.

JW-CH:
> Apologies - I did ask about it ImmichServerUrl in [#374 (comment)](https://github.com/immichFrame/ImmichFrame/pull/374#issuecomment-2960945554) and subsequently removed it.

Did not see that, no worries :)

JW-CH:
> Regarding the actual fix here, I'd suggest that the `ImmichServerUrl` is added to the dtos by the code which selects between the accounts to draw the assets from (in this case `TotalAccountImagesSelectionStrategy`). This way additional logic pools etc can be added or updated without needing to care about this, and the code to add the `ImmichServerUrl` (which comes from the config rather than the asset itself) is only in a single place, which is already a class that is concerned with the links between accounts and assets.

Not sure if I fully understand you. But the URL (or some reference) needs to be on the asset in the client to know where to navigate to.

jpg0:
The class `TotalAccountImagesSelectionStrategy` is responsible for selecting assets across multiple accounts. When asked for assets it pulls them from the various accounts (based on the total in each account). I am saying that when the assets are pulled from a specific account, the asset dtos can be "enriched" with the `ImmichServerUrl` for the account that they are pulled from, as they are pulled. This way it's all done in a single place which is already responsible for handling this mapping. In fact, because this class attempts to remember where the assets are from (using the bloom filters) it already handles the assets that it pulls individually, via calls to `_tracker.RecordAssetLocation(..)` on lines 24 & 69. It would be possible to set the `ImmichServerUrl` on the dtos at the same points.

jpg0:
I've implemented this in a PR here: https://github.com/immichFrame/ImmichFrame/pull/424

I don't mind which one you choose to merge!

PR #423, opened 2025-06-30
Chore: Switched to IMemoryCache to correctly handle expiration

jpg0:
Switched out the IDictionary cache implementation to an IMemoryCache. 

This allows for proper expiration; the previous one relied on attempting to retrieve an entry to trigger expiration. This one will expire and remove entries asynchronously. It's also a Microsoft class so less code for us to maintain.

(We could probably remove the ApiCache entirely and just use IMemoryCache directly, however the former uses per-cache expiration settings, whereas the latter is per-entry, so is not a drop-in replacement.)


JW-CH (approved):
LGTM :)

PR #424, opened 2025-06-30
Added ImmichServerUrl to AssetResponseDto

jpg0:
This is an alternative fix for https://github.com/immichFrame/ImmichFrame/pull/416

It works a little differently; the ImmichServerUrl is appended to the DTO by the class that is selecting across accounts, rather than the source pool. This allows future pool implementations not to care about the Url as it's appended later.

### DIFF
--- a/ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
+++ b/ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
@@ -6,7 +6,7 @@ using ImmichFrame.Core.Logic.Pool;
 
 namespace ImmichFrame.Core.Logic;
 
-public class PooledImmichFrameLogic : IImmichFrameLogic
+public class PooledImmichFrameLogic : IAccountImmichFrameLogic
 {
     private readonly IGeneralSettings _generalSettings;
     private readonly ApiCache _apiCache;
@@ -19,11 +19,15 @@ public class PooledImmichFrameLogic : IImmichFrameLogic
         _generalSettings = generalSettings;
 
         var httpClient = new HttpClient();
+
+        AccountSettings = accountSettings;
         httpClient.UseApiKey(accountSettings.ApiKey);
         _immichApi = new ImmichApi(accountSettings.ImmichServerUrl, httpClient);
         _apiCache = new ApiCache(TimeSpan.FromHours(generalSettings.RefreshAlbumPeopleInterval));
         _pool = BuildPool(accountSettings);
     }
+
+    public IAccountSettings AccountSettings { get; }
 
     private IAssetPool BuildPool(IAccountSettings accountSettings)
     {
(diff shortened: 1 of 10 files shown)


ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
```diff
@@ -25,6 +26,8 @@ public PooledImmichFrameLogic(IAccountSettings accountSettings, IGeneralSettings
         _pool = BuildPool(accountSettings);
     }
 
+    public IAccountSettings AccountSettings { get; }
```
JW-CH, line 30: This is never set and always null, no? The code does not work for me.

JW-CH (changes requested):

ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
jpg0: Apologies, you're completely right. I didn't get a chance to test against a real Immich server (as I was on a plane). I've updated the code to set this now and tested against a real server.

(This validates my PR to add testing without a real server: https://github.com/immichFrame/ImmichFrame/pull/425!)

JW-CH:
Could you also add the client side code in svelte? See my PR #416

jpg0:
Done

JW-CH:
There is a mismatch between the names `immichServerUrl` (Server) and `immichSourceUrl` (Client).

immichFrame.Web/src/lib/components/elements/imageoverlay/image-overlay.svelte
```diff
@@ -83,8 +83,7 @@
 				{/if}
 			</div>
 
-			<!-- TODO: Fix -->
-			<OverlayQr baseUrl={''} id={asset.id} />
+			<OverlayQr baseUrl={asset.immichSourceUrl ?? ''} id={asset.id} />
```
JW-CH:
```suggestion
			<OverlayQr baseUrl={asset.ImmichServerUrl ?? ''} id={asset.id} />
```

immichFrame.Web/src/lib/immichFrameApi.ts
```diff
@@ -103,6 +103,7 @@ export type TagResponseDto = {
 export type AssetTypeEnum = 0 | 1 | 2 | 3;
 export type AssetVisibility = 0 | 1 | 2 | 3;
 export type AssetResponseDto = {
+    immichSourceUrl?: string | null;
```
JW-CH:
```suggestion
    ImmichServerUrl?: string | null;
```

JW-CH:
Hello, me again, sorry :P `GetAssetInfoById()` also needs to return the new `ImmichServerUrl`.

jpg0:
Sorry again! Just updated.

PR #429, opened 2025-07-08
Allow overlays to be moved

Clete2:
One of the picture frames I maintain has limited line of sight to the bottom of the frame. I needed to be able to move all the overlays to the top. This feature allows repositioning of the main overlays in the project.

image

Each option accepts a vertical position (top, bottom) and a horizontal position (left, center, right).

From the updated documentation:

### Overlay Positioning
You can customize the position of different overlays on the screen:

- **ClockPosition**: Controls where the clock, date, and weather information appears
- **AssetInfoPosition**: Controls where image metadata (date, description, people, albums, location) appears
- **AppointmentsPosition**: Controls where calendar appointments appear

### DIFF
--- a/ImmichFrame.Core/Interfaces/IServerSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IServerSettings.cs
@@ -55,5 +55,8 @@
         public bool ImageFill { get; }
         public string Layout { get; }
         public string Language { get; }
+        public string ClockPosition { get; }
+        public string AssetInfoPosition { get; }
+        public string AppointmentsPosition { get; }
     }
 }
--- a/ImmichFrame.WebApi/Helpers/Config/ServerSettingsV1.cs
+++ b/ImmichFrame.WebApi/Helpers/Config/ServerSettingsV1.cs
@@ -111,5 +111,8 @@ public class ServerSettingsV1Adapter(ServerSettingsV1 _delegate) : IServerSettin
         public bool ImageFill => _delegate.ImageFill;
         public string Layout => _delegate.Layout;
         public string Language => _delegate.Language;
+        public string ClockPosition => "bottom-left";
+        public string AssetInfoPosition => "bottom-right";
+        public string AppointmentsPosition => "top-right";
     }
 }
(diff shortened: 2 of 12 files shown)


JW-CH:
This is already available with minimal effort via custom CSS. I am not really sure if this is something really needed. cc @3rob3

Clete2:
@JW-CH I saw that option too, but was thinking this would be an easier add. I am OK either way.

JW-CH:
> @JW-CH I saw that option too, but was thinking this would be an easier add. I am OK either way.

Don't get me wrong, I think an option in the settings would be easier for sure, but also more to maintain etc. Also, I just saw that some styles have rounded corners, you did not take this into account, right? (See Image)

image

3rob3:
Yeah, I was going to ask you the same thing. I agree this is far easier, but also increases the amount of settings. I'm also pretty sure I made a PR a while ago that did same thing, but we decided custom CSS was good enough and it was rejected. 
Again, I like this as it makes it much easier, but do we want to increases the amount of settings? I guess it's only 3.

ImmichFrame.Core/Interfaces/IServerSettings.cs
```diff
@@ -55,5 +55,8 @@ public interface IGeneralSettings
         public bool ImageFill { get; }
         public string Layout { get; }
         public string Language { get; }
+        public string ClockPosition { get; }
```
3rob3, line 58: I think if we move forward this should be something like "ClockWeatherPosition" (since you can show clock, weather, or both).

JW-CH:
> Yeah, I was going to ask you the same thing. I agree this is far easier, but also increases the amount of settings. I'm also pretty sure I made a PR a while ago that did same thing, but we decided custom CSS was good enough and it was rejected. Again, I like this as it makes it much easier, but do we want to increases the amount of settings? I guess it's only 3.

as mentioned, it is not only the amount of settings, also the lines in the client for the style according to the position (see the image I posted). I'd rather not do this here.

ImmichFrame.Core/Interfaces/IServerSettings.cs
Clete2: I'll let you two decide about the overall feature, if we do want to do this, I will rename.

PR #439, opened 2025-07-23
docs: fix Desktop section level and move keyboard section to Android

romainreignier:
The Desktop section was under the Android section.

But I moved the Keyboard section under Android because I could not find any keyboard handling in the Desktop app source code while the arrow keys are handled in the Android app here:

https://github.com/immichFrame/ImmichFrame_Android/blob/8eb0ac8a6e2e4b583eb58418fd2083b072329a9b/app/src/main/java/com/immichframe/immichframe/MainActivity.kt#L704-L737

But I could not find the Arrow Down handling to quit the application :(

### DIFF
--- a/docs/docs/getting-started/apps.md
+++ b/docs/docs/getting-started/apps.md
@@ -140,7 +140,15 @@ Follow instructions below to update WebView to 106. This has been tested and wor
   ```shell
   adb reboot
   ```
-### Desktop (Windows, MacOS, Linux)
+
+### Keyboard
+**Settings** - Up arrow  
+**Quit** - Down arrow  
+**Prev Image** - Left arrow  
+**Next Image** - Right arrow  
+**Pause** - Enter/Return  
+
+## Desktop (Windows, MacOS, Linux)
 
 Get the latest Desktop-App from the [GitHub Releases][releases-url]-Page.
 
@@ -158,14 +166,6 @@ The screen is configured in a 3x3 gird. You can touch or click:
   ```bash 
   xattr -c /Applications/immichframe.app
   ```
-
-## Keyboard
-**Settings** - Up arrow  
-**Quit** - Down arrow  
-**Prev Image** - Left arrow  
-**Next Image** - Right arrow  
-**Pause** - Enter/Return  
-
 
 <!-- MARKDOWN LINKS & IMAGES -->
 [play-store-link]: https://play.google.com/store/apps/details?id=com.immichframe.immichframe


3rob3:
I can’t check right now, but my memory says it is in the Rust code.

romainreignier:
I don't think so.
I have manage to add some but only in the javascript [on my fork](https://github.com/romainreignier/ImmichFrame_Desktop/commit/0175b8b247eb65b7f968eaa0b4258ebfb48e5d67).

3rob3:
Yeah you’re right, I was thinking about the exit_app method. Arrow left right may work since the “WebView” receives them, but settings and exit are click/touch only.

romainreignier:
Yes, this is almost this, see https://github.com/immichFrame/ImmichFrame_Desktop/issues/6#issue-3256671661

Except that the iframe has not the focus by default.

I have added:
```js
iframe.contentWindow.focus();
```

[in the javascript](https://github.com/immichFrame/ImmichFrame_Desktop/blob/52c1ec5fefc2c70da482f92c974bc3be624a71c9/src/main.js#L18) and now it works without clicking.

Edit: I have pushed my changes [here](https://github.com/romainreignier/ImmichFrame_Desktop/commit/4b84190a4883b026c882630b157d6185b49df915).

3rob3:
Can you move Keyboard back under Desktop? This will be fixed/re-implemented soon.

3rob3:
> I don't think so. I have manage to add some but only in the javascript [on my fork](https://github.com/romainreignier/ImmichFrame_Desktop/commit/0175b8b247eb65b7f968eaa0b4258ebfb48e5d67).

Nice, does this interfere with arrow left/right (which should be forwarding through to the web app)? If not do want to to create a PR for this?

romainreignier:
> Can you move Keyboard back under Desktop? This will be fixed/re-implemented soon.

Only under Desktop or on a higher level to make it common?

I don't like to release documentation on features not yet released. It was quite confusing to me as a new comer (YAML settings format for example).

> Nice, does this interfere with arrow left/right (which should be forwarding through to the web app)?

Yes it interferes. The focus is either in the Tauri app or the Webview but cannot be both.
Until now, by default the focus is is on Tauri so my changes work to open settings and quit. But we cannot use right / left.
If we click with a mouse, then the focus goes to the webview and then the right / left arrows start to work and the Up and Down do not work anymore.
To set the focus on the webview by default I had to make [these changes](https://github.com/romainreignier/ImmichFrame_Desktop/commit/4b84190a4883b026c882630b157d6185b49df915).

3rob3:
> > Can you move Keyboard back under Desktop? This will be fixed/re-implemented soon.
> 
> Only under Desktop or on a higher level to make it common?
> 
> I don't like to release documentation on features not yet released. It was quite confusing to me as a new comer (YAML settings format for example).
> 
> > Nice, does this interfere with arrow left/right (which should be forwarding through to the web app)?
> 
> Yes it interferes. The focus is either in the Tauri app or the Webview but cannot be both. Until now, by default the focus is is on Tauri so my changes work to open settings and quit. But we cannot use right / left. If we click with a mouse, then the focus goes to the webview and then the right / left arrows start to work and the Up and Down do not work anymore. To set the focus on the webview by default I had to make [these changes](https://github.com/romainreignier/ImmichFrame_Desktop/commit/4b84190a4883b026c882630b157d6185b49df915).

ugh, this is bringing back memories (and why I never re-implemented keyboard control). I remember fighting this issue. Not sure what we should do about keyboard section then, I guess just leave it like you have it.

PR #441, opened 2025-07-24
Permit RefreshAlbumPeopleInterval of 0

jpg0:
Allow RefreshAlbumPeopleInterval to be set to zero, which will set the cache timeout to 1ms (it cannot be set to TimeSpan.Zero).


3rob3:
Nice job and thanks for the quick turnaround.

PR #442, opened 2025-07-25
Fix: Memories cache should last one day

jpg0:
I noticed that the API caching is set to the same for all types of asset, however it should be fixed to one day for memories as they should be refreshed daily.

### DIFF
--- a/ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
+++ b/ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
@@ -42,7 +42,7 @@ public class PooledImmichFrameLogic : IAccountImmichFrameLogic
             pools.Add(new FavoriteAssetsPool(_apiCache, _immichApi, accountSettings));
 
         if (accountSettings.ShowMemories)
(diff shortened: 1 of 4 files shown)


ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
```diff
@@ -42,7 +42,7 @@ private IAssetPool BuildPool(IAccountSettings accountSettings)
             pools.Add(new FavoriteAssetsPool(_apiCache, _immichApi, accountSettings));
 
         if (accountSettings.ShowMemories)
-            pools.Add(new MemoryAssetsPool(_apiCache, _immichApi, accountSettings));
+            pools.Add(new MemoryAssetsPool(new ApiCache(TimeSpan.FromDays(1)), _immichApi, accountSettings));
```
JW-CH: Shouldn't this be cached until 00:00?

jpg0: I guess that is a little better - I've updated it to do that instead.

JW-CH:
@jpg0 tests failed, could you check this? :)

jpg0:
Fixed

PR #447, opened 2025-08-02
doc: add instruction to find frameo webview path

mammuth:
My Denver Frameo model has the webview placed inside `/product/app/webview`, not `/system`. 🤷 

This may help others to apply the changes regardless of the model they have.

### DIFF
--- a/docs/docs/getting-started/apps.md
+++ b/docs/docs/getting-started/apps.md
@@ -116,9 +116,13 @@ Follow instructions below to update WebView to 106. This has been tested and wor
   ```shell
   adb push /path/to/your/new/webview.apk /sdcard/
   ```
-- Enter root shell
+- Enter shell and switch to the root user
   ```shell
-  adb shell su
+  adb shell
+  ```
+  then inside the shell
+  ```shell
+  su
   ```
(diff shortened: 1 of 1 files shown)


docs/docs/getting-started/apps.md
```diff
@@ -116,9 +116,13 @@ Follow instructions below to update WebView to 106. This has been tested and wor
   ```shell
   adb push /path/to/your/new/webview.apk /sdcard/
   ```
-- Enter root shell
+- Enter shell and switch to the root user
```
3rob3, line 119: Why break this up into two commands when the one liner works and is simpler?  Did you have trouble?

3rob3: Otherwise looks good, thanks for contributing!

mammuth: > Did you have trouble?

Yes, I was also surprised by this but I received error messages (that I don't exactly recall, but was something about "command `su` timed out" or something like that).

But I also only had a windows PC available when I tried it, so maybe it was related to the Windows Android SDK or the way the Windows cmd works. No clue 🤷

PR #455, opened 2025-08-26
Creating new config property for clock date format (ClockDateFormat)

jtlawren67:
I wanted to having the Date above the clock be in a separate format from the Photo Date so I created a new property in the Config for ClockDateFormat and altered the clock.svelte file to use the new ClockDateFormat as opposed to PhotoDateFormat.

This is my first time ever touching svelte or C# so mostly hacking around.  

Would love to know any tips for how to better help develop.

### DIFF
--- a/ImmichFrame.Core/Interfaces/IServerSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IServerSettings.cs
@@ -37,6 +37,7 @@
         public int RenewImagesDuration { get; }
         public bool ShowClock { get; }
         public string? ClockFormat { get; }
+		public string? ClockDateFormat { get; }
         public bool ShowProgressBar { get; }
         public bool ShowPhotoDate { get; }
         public string? PhotoDateFormat { get; }
--- a/ImmichFrame.WebApi/Helpers/Config/ServerSettingsV1.cs
+++ b/ImmichFrame.WebApi/Helpers/Config/ServerSettingsV1.cs
@@ -33,6 +33,7 @@ public class ServerSettingsV1 : IConfigSettable
     public double TransitionDuration { get; set; } = 1;
     public bool ShowClock { get; set; } = true;
     public string? ClockFormat { get; set; } = "hh:mm";
+	public string? ClockDateFormat { get; set; } = "MM/dd/yyyy";
     public bool ShowProgressBar { get; set; } = true;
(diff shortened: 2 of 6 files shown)


JW-CH:
Thank you for your contribution.

Could you explain why you would like to be the `ClockDateFormat` not the same as the `PhotoDateFormat`? I don't really see a use-case for that. Having two formats for those dates creates more initial effort to configure immichFrame.

jtlawren67:
Sure, the very direct use case is that I view the clock as filling a
calendar job where I would like to know the day of week (i forget this more
than I'd like).  However, for the photo itself I don't really care if a
picture from 4 years ago was on a Wednesday so having the day of week for
only the clock was valuable.

That said, your point is valid about having excess configurations.
Originally I thought that could be mitigated by having the clockDateFormat
default to the photoDateFormat if not populated but wasn't sure how to do
that.

I'm forking to try to add a few features that I had on MagicMirror for a
personal use-case and some of those are WAY too specific to try to
contribute to the main package.  But for this one I figured that it was
simple enough that someone else might have the need.

No worries if you'd rather not add it.  Thanks for replying and for
building everything you have that allowed me to even do this simple piece!

3rob3:
I'm not against it. Way back (Avalonia days) ClockFormat used to include the time AND date in the format. I started a PR a while ago to re-implement it that way but it was difficult with svelte to keep them on separate lines with all the different possible regional formats. This is a decent compromise. I'd have to update the non-webview clients though.....

Also, you need to also update the example configs in the docker folder (env, json, yml).

3rob3:
I’m good with it, @JW-CH you?

JW-CH:
> I’m good with it, @JW-CH you?

Tests are failing. @jtlawren67 can you fix them? You have to add them to the yaml/json in the test project.

jtlawren67:
yup, will fix the test in the next couple days

3rob3:
#467 is nearly identical to this, but also includes the AssetController test case. Do you mind if I merge that one instead?

3rob3:
Superseded by #467

jtlawren67:
yeah its okay.  I understand why you'd go with the other one.

3rob3:
Hope you’re not dissuaded from contributing. We appreciate it!

PR #467, opened 2025-09-19
Add ClockDateFormat string to format the date on the clock

rgregg:
Add support for custom date format for clock which is separate from the date display on assets/images. This allows you to use a more "now" relative format - like "Tues, Sept 18" for the clock when you still want to be able to show the day/month/year for assets.

Sets the default to that format "eee, MMM d" which is a change from previous default value.


rgregg:
I just realized this is the same as #455 - oops!

3rob3:
Yes, looks nearly identical to #455 :). Main difference is this includes AssetController tests. @JW-CH , do you have a preference?

immichFrame.Web/src/lib/components/elements/appointments.svelte
```diff
@@ -13,8 +13,8 @@
 		let sameDay = startDate.getDate() == endDate.getDate();
 
 		let clockFormat = $configStore.clockFormat ?? 'HH:mm';
-		let photoDateFormat = $configStore.photoDateFormat ?? 'dd.MM.yyyy';
-		let fullFormat = photoDateFormat + ' ' + clockFormat;
+		let clockDateFormat = $configStore.clockDateFormat ?? 'eee, MMM d';
```
JW-CH, line 16: We could debate if the date in the appointments should apply the `clockDateFormat`.

JW-CH:
> Yes, looks nearly identical to #455 :). Main difference is this includes AssetController tests. @JW-CH , do you have a preference?

Oops, sorry. Did not see this. I don't mind which one to merge. I prefer this one because this also updates the docs. Even though the other pull request had already been opened beforehand.

PR #468, opened 2025-09-19
Add the weather icon for current conditions

rgregg:
This update adds displaying the weather icon for the current conditions via a configural URL in the settings file. If the value is not left empty no icon is displayed. 

The default is to use the icon from the OpenWeatherMap service - but you can specify any URL that will return the right content - including caching the icons locally. `{IconId}` is a placeholder in the template replaced with the value of the IconId as returned by the IWeather instance.

This PR also updates the CSS to allow for richer customizations of the date, clock, and weather via custom CSS.


3rob3:
Looks good, I like it! We used to have this (then we had some custom icons) in the old Avalonia standalone version. I like that this allows for both (or none).

3rob3:
Do you have examples of other URL's you can use?

rgregg:
Nothing out of the box - the information for OpenWeatherMap icons is here: https://openweathermap.org/weather-conditions#How-to-get-icon-URL

My thought with making this configurable is that someone might want to source the icons locally, so you could change the URL to somethings served off the ImmichFrame static site and create a copy of the icons there or add your own custom ones: `http://localhost:8080/static/weather/{IconId}.png`

If in the future there are other weather providers - this also allows the icon URL to be easily updated for alternative weather provide implementation.

PR #487, opened 2025-10-10
Fix: Missing config directory

JW-CH:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration file discovery when the specified directory is missing. The app now safely falls back to a default filename without attempting to enumerate files, preventing startup errors and improving reliability on first run or in clean environments.


JW-CH:
cc @3rob3 merging this. The `FindConfigFile()` failed when ENV vars were used and /app/config was not existing.

3rob3:
Agree, thanks

PR #490, opened 2025-10-11
Feat: Update QR Code to redirect to my.immich.app

### LINKED ISSUE #482: QR Code should open links at my.immich.app instead of the user's server url
### Version

1.0.28.0

### Deployment Method

Android

### Description

When scanning the QR Code, it should send users to `https://my.immich.app/[relevant photo url]`, This url can be used to open the photo directly on the mobile App, something that's not possible with the current setup of opening the user's server URL. It also allows users to setup the redirect URL on desktop envs.

### Reproduction

1. Scan the QR Code with a phone that has the Immich app
2. The QR code sends the user to Immich on the browser instead of the mobile app.

### Expectations

Should open the mobile app on the relevant photo

### Configuration

```xml

```

### Logs

```shell

```

### Pre-Submission Checklist

- [x] This is a bug report and not a feature request

- [x] I have provided all of the required information to reproduce the bug (config, logs, etc.)

leocb:
Fixes #482

This PR redirects users scanning the QR Code to `my.immich.app` - This is the recommended URL to call when you want to open Immich URLs on your local instance, it also open the mobile app when it's installed. Here's a video of it working: 

https://github.com/user-attachments/assets/dc4bdbc9-d91a-488c-ab25-f447be7e488b

## Summary by CodeRabbit

- Bug Fixes
  - QR overlay now reliably displays using the app’s server address, preventing failures caused by missing or incorrect base URLs.
- Refactor
  - Simplified QR overlay configuration by removing the need to provide a server URL.
  - Reduced setup complexity for image overlays while keeping existing behavior intact.

### DIFF
--- a/ImmichFrame.Core.Tests/ImmichFrame.Core.Tests.csproj
+++ b/ImmichFrame.Core.Tests/ImmichFrame.Core.Tests.csproj
@@ -9,19 +9,19 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="9.0.6" />
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
-        <PackageReference Include="NUnit" Version="4.1.0" />
-        <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
-        <PackageReference Include="NUnit.Analyzers" Version="4.2.0">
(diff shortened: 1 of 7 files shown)


JW-CH:
Thank you for your PR. I just thought a bit more about this. This will only work if you are using a single instance of immich. In theory, it would be possible to have immichframe pull image from multiple servers. This then will only work with your configured server. I think most people will only have one immich server they pull images from, so this is pretty nieche. LGTM

Issue #509, opened 2025-11-20
Feature request: configurable "config" directory


jfly:
### Description

Right now, ImmichFrame searches for a "Config" directory relative to `AppDomain.CurrentDomain.BaseDirectory`: <https://github.com/immichFrame/immichFrame/blob/6ab6554478a22140fd5accd8bbceb7d9b61b1134/ImmichFrame.WebApi/Program.cs#L49-L51>.

I'd like to be able to choose a different directory, either via an environment variable, or a command line flag.

### Use Case

I'm packaging ImmichFrame for nix/nixos, and we don't tend to put our application config right next to the applications themselves. I'm pretty sure this is standard for all package managers/distros.

### Pre-Submission Checklist

- [x] This is a feature request and not a bug report

- [x] I have checked for related issues and checked the documentation

PR #512, opened 2025-11-20
Configurable config path

jfly:
This fixes https://github.com/immichFrame/ImmichFrame/issues/509

Edit: I did not write the AI generated summary below. It makes this change sound more complicated than it is.

## Summary by CodeRabbit

* **Chores**
  * Config discovery now supports an environment variable to specify a custom config path, with improved detection and reliable fallbacks.

* **Documentation**
  * Expanded configuration guidance including config lookup order and detailed defaults; fixed a malformed code fence in customization docs.

* **Bug Fixes**
  * Fixed CSS syntax and list-style rules to ensure consistent styling.

### DIFF
--- a/ImmichFrame.WebApi/Program.cs
+++ b/ImmichFrame.WebApi/Program.cs
@@ -46,7 +46,8 @@ builder.Services.AddLogging(builder =>
 
 
 // Setup Config
-var configPath = Directory.EnumerateDirectories(AppDomain.CurrentDomain.BaseDirectory, "*", SearchOption.TopDirectoryOnly)
+var configPath = Environment.GetEnvironmentVariable("IMMICHFRAME_CONFIG_PATH") ??
+        Directory.EnumerateDirectories(AppDomain.CurrentDomain.BaseDirectory, "*", SearchOption.TopDirectoryOnly)
         .FirstOrDefault(d => string.Equals(Path.GetFileName(d), "Config", StringComparison.OrdinalIgnoreCase))
         ?? Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config");
 builder.Services.AddTransient<ConfigLoader>();
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -43,3 +43,10 @@
   height: 0.9em;
   width: 0.9em;
 }
+
+/*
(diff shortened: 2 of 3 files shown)


ImmichFrame.WebApi/Program.cs
```diff
@@ -46,7 +46,8 @@ _ _ __ ___  _ __ ___  _  ___| |__ | |_ _ __ __ _ _ __ ___   ___
 
 
 // Setup Config
-var configPath = Directory.EnumerateDirectories(AppDomain.CurrentDomain.BaseDirectory, "*", SearchOption.TopDirectoryOnly)
+var configPath = Environment.GetEnvironmentVariable("CONFIG_PATH") ??
+        Directory.EnumerateDirectories(AppDomain.CurrentDomain.BaseDirectory, "*", SearchOption.TopDirectoryOnly)
         .FirstOrDefault(d => string.Equals(Path.GetFileName(d), "Config", StringComparison.OrdinalIgnoreCase))
         ?? Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config");
```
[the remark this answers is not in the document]
jfly: > Missing CONFIG_PATH validation

I disagree with this concern: if the folder doesn't exist, let's just see what happens when we try to read it, we should get a useful error message, and if we don't, that's room for future improvement.

The other concerns are largely complaining about pre-existing issues.

JW-CH:
Thank you for your PR! Could you also add some docs for the new env variable?

jfly:
> Could you also add some docs for the new env variable?

Done!

docs/src/css/custom.css
```diff
@@ -42,4 +42,11 @@
...
-}
\ No newline at end of file
+}
+
+/*
+ * Override Tailwind's [default unstyled lists](https://github.com/tailwindlabs/tailwindcss/blob/v4.1.17/packages/tailwindcss/preflight.css#L193-L201).
+ */
+.markdown li {
+  list-style-type: decimal;
+}
```
jfly, line 52: Alternatively, could we remove Tailwind? I'm not sure what we're using it for here, and I'm not sure it plays nicely with Docusaurus's own opinionated css framework (Infima)

JW-CH: I am totally fine with removing Tailwind. I did not really recall adding it. :P I was orienting myself at the immich docs, and they added Tailwind too on their end. My suggestion would be to do it in a separate PR, only the landing-page is using tailwind.

jfly:
Thanks for the merge, @JW-CH!

Issue #523, opened 2025-12-11
Weather Icon Images Fail to Display when Multiple Icons are called (eg: 'light Intensity drizzle, mist')


andyk-immich:
### Version

1.0.30

### Deployment Method

Web

### Description

When the source from openweathermap.org for my lat-long has two weather icon images (eg: moderate rain, mist), it fails to display the images and distorts the layout.

The image linked in the source appears to link to:

openweathermap.org/img/wn/10n.png <- works with one image
openweathermap.org/img/wn/10n,50n.png  <- does not work with two or greater icon images comma separated

![Image](https://github.com/user-attachments/assets/19c56215-0332-4685-950a-69237878dabc)

![Image](https://github.com/user-attachments/assets/88b70509-1aa1-4232-9542-e53ee94172c1)

### Reproduction

If the weather for a specific area has two or more descriptors, two or more images will attempt to link in url and this fails.  

### Expectations

Either multiple icons are ignored and only one icon image appears, or multiple appear ok.

### Configuration

```xml
name: immichframe1
services:
  immichframe:
    container_name: immichframe1
    image: ghcr.io/immichframe/immichframe:latest
    restart: on-failure
    ports: 
      - 30042:8080
    environment:
      TZ: America/Vancouver
      ImmichServerUrl: http://<redacted>
      ApiKey: <redacted>
      #AuthenticationSecret: ""
      Interval: 10
      TransitionDuration: 2
      ImageZoom: false
      #ImageFill: "false"
      #Layout: splitview         
      #DownloadImages: "false"
      #ShowMemories: "false"
      #ShowFavorites: "false"
      #ShowArchived: "false"
      #ImagesFromDays: ""
      #ImagesFromDate: ""
      #ImagesUntilDate: ""
      #RenewImagesDuration: "30"
      #Rating: "5"
      Albums: <redacted>
      #ExcludedAlbums: "ALBUM3,ALBUM4"
      #People: "PERSON1,PERSON2"
      #Webcalendars: "https://calendar.mycalendar.com/basic..."
      #RefreshAlbumPeopleInterval: "12"
      ShowClock: true
      ClockFormat: HH:mm
      ShowPhotoDate: true
      PhotoDateFormat: yyyy-MM-dd
      #ShowImageDesc: "true"
      #ShowPeopleDesc: true
      ShowImageLocation: true
      ImageLocationFormat: City
      ShowPeopleDesc: false
      ShowAlbumName: false
      ShowProgressBar: false
[...]

PR #526, opened 2025-12-11
When Openweathermap calls for multiple images, split icon Id and get multiple images

TomNaughton:
[Issue link](https://github.com/immichFrame/ImmichFrame/issues/523)

iconId is split at comma and then iterated over, producing icons per iconId.

## Summary by CodeRabbit

* **New Features**
  * The weather icon rendering on the clock widget has been enhanced to display multiple weather condition icons simultaneously. When multiple weather conditions are detected, each condition is now shown with its own corresponding icon, allowing users to gain a comprehensive understanding of current atmospheric conditions at a glance, rather than seeing only a single consolidated weather representation.

### DIFF
--- a/immichFrame.Web/src/lib/components/elements/clock.svelte
+++ b/immichFrame.Web/src/lib/components/elements/clock.svelte
@@ -75,8 +75,14 @@
 				id="clockweatherinfo"
 				class="text-xl sm:text-xl md:text-2xl lg:text-3xl font-semibold text-shadow-sm weather-info"
 			>
-				{#if $configStore.weatherIconUrl }
-				<img src="{ $configStore.weatherIconUrl.replace('{IconId}', encodeURIComponent(weather.iconId)) }" class="icon-weather" alt="{weather.description}">
+				{#if $configStore.weatherIconUrl}
+					{#each weather.iconId?.split(',') ?? [] as icon (icon)}
+						<img
+							src={$configStore.weatherIconUrl.replace('{IconId}', encodeURIComponent(icon.trim()))}
+							class="icon-weather"
+							alt={weather.description}
+						/>
+					{/each}
 				{/if}
 				<div class="weather-location">{weather.location},</div>
 				<div class="weather-temperature">{weather.temperature?.toFixed(1)}</div>


immichFrame.Web/src/lib/components/elements/clock.svelte
```diff
@@ -75,8 +75,14 @@
 				id="clockweatherinfo"
 				class="text-xl sm:text-xl md:text-2xl lg:text-3xl font-semibold text-shadow-sm weather-info"
 			>
-				{#if $configStore.weatherIconUrl }
-				<img src="{ $configStore.weatherIconUrl.replace('{IconId}', encodeURIComponent(weather.iconId)) }" class="icon-weather" alt="{weather.description}">
+				{#if $configStore.weatherIconUrl}
+					{#each weather.iconId?.split(',') ?? [] as icon (icon)}
```
3rob3, line 79: Have you tested this and it looks OK? I feel like it would look weird having two weather icons. My vote would be split iconId on comma, and just always use element [0].

JW-CH: I was thinking the same. Also, if the `weather.iconId` would be an empty string, the split would still return a single element (empty) which would request the URL with an empty `id`.

TomNaughton: I considered displaying a single element. This may work fine, but it may result in discounting a weather status that the user would prefer to know about. If the API returns "mist,heavy rain", the second element should probably take priority.

3rob3: Seems like a few people are seeing this issue and we should merge this. Still not sure about whether we should show all icons or just one.

CRhodey: > Seems like a few people are seeing this issue and we should merge this. Still not sure about whether we should show all icons or just one.

As Tom said above, I think my preference would be to go for the dominant or highest impact weather. I.E: Thunderstorm > Freezing rain > Snow > Hail > Sleet > Rain > Drizzle

3rob3: Then I think my original suggestion still applies then (just use element [0]):
https://openweathermap.org/weather-conditions
_NOTE: It is possible to meet more than one weather condition for a requested location. The first weather condition in API respond is primary._

JW-CH: > Then I think my original suggestion still applies then (just use element [0]): https://openweathermap.org/weather-conditions _NOTE: It is possible to meet more than one weather condition for a requested location. The first weather condition in API respond is primary._

Also note my reply, as of now, the code does not work good.

3rob3: > > Then I think my original suggestion still applies then (just use element [0]): https://openweathermap.org/weather-conditions _NOTE: It is possible to meet more than one weather condition for a requested location. The first weather condition in API respond is primary._
> 
> Also note my reply, as of now, the code does not work good.

I submitted #569 as an alternative to this

Issue #305, opened 2025-04-12
Progress bar hidden by white bar at bottom of iPad screen


simondsmason:
### Version

latest on 4/12/25. I don't know how to find the version number when using :latest?

### Deployment Method

Unraid docker - Janmer's repository

### Description

Running full screen on an iPad Pro (12.9 inch) running Version 16.7.10 IOS.  Instead of the progress bar it displays a white bar that runs across the entire width of the iPad.

### Reproduction

Open immichframe in safari.  Save to desktop.  Open the saved icon.  If you swipe down and hold the bar goes down and shows the progress bar.  If you let go it will come up again and block the progress bar.

### Expectations

There should not be a white bar - only the visible progress bar.

### Configuration

```xml
I don't see a copy of settings.xml.  I don't think this template uses it as I specify everything in the docker template?
```

### Logs

```shell
warn: Microsoft.AspNetCore.DataProtection.Repositories.EphemeralXmlRepository[50]
      Using an in-memory repository. Keys will not be persisted to storage.
warn: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[59]
      Neither user profile nor HKLM registry available. Using an ephemeral key repository. Protected data will be unavailable when application exits.
warn: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[35]
      No XML encryptor configured. Key {a72a4229-cfc4-48b5-a2d5-d858b265bfab} may be persisted to storage in unencrypted form.
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://[::]:8080
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Production
info: Microsoft.Hosting.Lifetime[0]
      Content root path: /app
```

### Pre-Submission Checklist

- [x] This is a bug report and not a feature request

- [x] I have provided all of the required information to reproduce the bug (config, logs, etc.)

- [x] I have checked for related issues and checked the documentation

3rob3:
Hi. I’m pretty sure this PR addresses the same issue:
https://github.com/immichFrame/ImmichFrame/pull/287
We should have a release soon that includes this.

teedoff087:
Hey ImmichFrame Team,

[...]

PR #528, opened 2025-12-16
Fix: Black background for iOS safe area fix

3rob3:
Addresses new comment on #305 

## Summary by CodeRabbit

* **Style**
  * Updated page background color to black.

### DIFF
--- a/immichFrame.Web/src/app.css
+++ b/immichFrame.Web/src/app.css
@@ -57,6 +57,7 @@
 }
 
 html {
+	background-color: black;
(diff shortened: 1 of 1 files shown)


JW-CH:
Is this not handled by the OS? Setting a dark/light theme?

3rob3:
If it was a native app, yes, but not web apps. The 'color-scheme: light dark' is honored, but just for inputs, scrollbars, etc (not for the backing layer). This is a weird iOS quirk.

PR #537, opened 2025-12-30
Feat: support selecting photos from tags

ahal:
Personally I'd prefer to use tags to select which photos show up on my frame, rather than adding photos to an album.

## Summary by CodeRabbit

* **New Features**
  * Per-account tag filtering and a new tag-based asset collection; tag collections appear alongside existing categories.
  * New "Show Tags Description" toggle to display comma-separated tag names in asset views, overlays, and client settings.

* **Documentation**
  * Configuration guide updated with tag setup, hierarchical/path matching examples, and required API permission.

* **Tests**
  * Added comprehensive unit tests plus minor test scaffolding improvements covering loading, pagination, matching, enrichment, and deduplication.

### DIFF
--- a/ImmichFrame.Core.Tests/Logic/Pool/TagAssetsPoolTests.cs
+++ b/ImmichFrame.Core.Tests/Logic/Pool/TagAssetsPoolTests.cs
@@ -72,6 +72,18 @@ public class TagAssetsPoolTests
     }
 
     [Test]
+    public async Task LoadAssets_NullTagsConfigured_ReturnsEmpty()
+    {
+        _settings.SetupGet(s => s.Tags).Returns(null as List<string>);
+        _api.Setup(a => a.GetAllTagsAsync(default)).ReturnsAsync([]);
+
+        var result = await _pool.LoadAssetsPublic();
+
+        Assert.That(result, Is.Empty);
+        _api.Verify(a => a.SearchAssetsAsync(It.IsAny<MetadataSearchDto>(), default), Times.Never);
+    }
+
+    [Test]
     public async Task LoadAssets_NoTagsConfigured_ReturnsEmpty()
     {
         _settings.SetupGet(s => s.Tags).Returns(new List<string>());
--- a/ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
+++ b/ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
@@ -8,6 +8,12 @@ public class TagAssetsPool(IApiCache apiCache, ImmichApi immichApi, IAccountSett
     protected override async Task<IEnumerable<AssetResponseDto>> LoadAssets(CancellationToken ct = default)
     {
         var tagAssets = new List<AssetResponseDto>();
+
+        if (accountSettings.Tags == null)
+        {
+            return tagAssets;
+        }
+
         var allTags = await apiCache.GetOrAddAsync($"allTags_{accountSettings.ImmichServerUrl}",
             () => immichApi.GetAllTagsAsync(ct));
         var tagValueToTag = allTags.ToDictionary(t => t.Value);


ahal:
This PR is AI assisted. I've reviewed and tested it locally.

docs/docs/getting-started/configuration.md
```diff
@@ -130,16 +132,21 @@ Accounts:
     # UUID of People
     People:  # string[]
       - UUID
+    # UUID of Tags
+    Tags:  # string[]
+      - UUID
```
JW-CH: What's the easiest way to get a tag UUID for a user? Wouldn't `value` be the better option?

ahal: Maybe? I guess there could be issues when renaming tags. Mostly was trying to stay consistent with people and albums. If we used value for tags, then why not use a value for people and albums as well?

But I'm happy to change it to value if you prefer

ahal: Re: easiest way, I didn't see a way to get it outside the API. But maybe I didn't look hard enough. Personally I'm fine with using the API, but you're right that for most users, that would be a barrier.

Probably worth mentioning that tags aren't enabled by default, so they're already power user territory.

JW-CH: > Maybe? I guess there could be issues when renaming tags. Mostly was trying to stay consistent with people and albums. If we used value for tags, then why not use a value for people and albums as well?
> 
> But I'm happy to change it to value if you prefer

People and Albums could have the same `value` but not be the same; when I checked, that was not the case for `tags`. I just saw that the `tags` in Immich does not display a UUID; that's why I asked.

With `albums` or `people`, users can easily get the UUID via the URI.

I'd suggest using the tags `value` and, in the backend, getting all tags and their `ID`s manually (https://api.immich.app/endpoints/tags/getAllTags)

Please let me know what you think about this. :) I know that it would be easier to use the UUID for you, but for the end user it's better to have the value, since there won't be tag duplicates (afaik)

ahal: Makes sense, and agreed!

I'll update the PR when I get a chance :)

ahal: One small thing I didn't realize is that `value` is the full path of a tag, whereas `name` is just the last part.. E.g, you could have:
```
- new-years/party
- birthday/party
```

The value is the full thing, but the name is simply `party`. I think specifying values makes the most sense, but there's a small gotcha where some folks might expect to be able to specify the `name` and be surprised when it doesn't work. I'll make sure there's mention of this in the docs.

As a follow-up it might be nice to support either `name` or `value`.. e.g perhaps it first looks for a match in names, and if none are found, it falls back to value.. But I'd rather not deal with that extra complexity at this time :)

ahal:
No rush, but just in case it wasn't clear, this should be ready for another round of review :)

docker/Settings.example.yml
```diff
@@ -52,3 +52,6 @@ Accounts:
       - UUID
     People:
       - UUID
+    Tags:
+      - Vacation
```
JW-CH, line 56: Could you update the Tags according to the documentation? I liked that you were showing a subtag there too! (`Travel/Europe`) Same for the `JSON` example.

JW-CH (commented):
Could you run `make api` in the console to generate the `swagger.json`. Make sure the app is running and do it in another console :) also add the topmost line again. If you don't know what I want I can also do it myself :P

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
```diff
@@ -0,0 +1,73 @@
...
+using ImmichFrame.Core.Interfaces;
+
+namespace ImmichFrame.Core.Logic.Pool;
+
+public class TagAssetsPool(IApiCache apiCache, ImmichApi immichApi, IAccountSettings accountSettings) : CachingApiAssetsPool(apiCache, immichApi, accountSettings)
+{
+    protected override async Task<IEnumerable<AssetResponseDto>> LoadAssets(CancellationToken ct = default)
+    {
+        var tagAssets = new List<AssetResponseDto>();
+        var allTags = await immichApi.GetAllTagsAsync(ct);
```
JW-CH: Can we cache the tags maybe?

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                // Fetch full asset details to get tag information
+                foreach (var asset in tagInfo.Assets.Items)
+                {
+                    if (seenIds.Contains(asset.Id))
+                    {
+                        continue;
+                    }
+
+                    if (asset.Tags == null)
```
JW-CH: Why is this checked for null? Tags should not be null here if you search like this no?

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
```diff
@@ -0,0 +1,73 @@
...
+            int page = 1;
+            int batchSize = 1000;
+            int itemsInPage;
+            do
+            {
+                var metadataBody = new MetadataSearchDto
+                {
+                    Page = page,
+                    Size = batchSize,
+                    TagIds = [tagId],
```
JW-CH: the search can take multiple tags, why looping over and just take one?

JW-CH: Just saw that the person pool does not do this too, but you could :)

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
```diff
@@ -0,0 +1,73 @@
...
+                };
+
+                var tagInfo = await immichApi.SearchAssetsAsync(metadataBody, ct);
+
+                itemsInPage = tagInfo.Assets.Items.Count;
+
+                // Fetch full asset details to get tag information
+                foreach (var asset in tagInfo.Assets.Items)
+                {
+                    if (seenIds.Contains(asset.Id))
```
JW-CH, line 56: Also tracking `seenIds` is not needed when using one search (would be my guess :P)

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
ahal: I think this is actually a bug in Immich:
https://github.com/immich-app/immich/discussions/17361

Even though we're filtering by tag id, the tags aren't being returned in the response of `SearchAssetsAsync`, so another call is needed to get them.

But I think this was written back when I was using tag ids in the configuraiton. Now that we're specifying the tag value directly, we don't really need to get all the tags and we could just display that. The caveat is that in this current implementation, all tags for an asset are displayed, but if we drop the call to `GetAssetInfoAsync`, we'd only display the configured tag.

Personally I think this is fine and makes sense, so I'll go ahead and make that change. If you disagree I can revert it back.

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
ahal: Given https://github.com/immichFrame/ImmichFrame/pull/537#discussion_r2692862080 I think I now actually do need to loop over them one by one. Otherwise there would be no way to know which tag to display as we aren't calling `GetAssetInfoAsync` anymore.

I'll publish this change soon and you can decide which one you prefer.

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
ahal: Bug might be the wrong word, seems more like `WithTags` just isn't implemented yet.

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
ahal: I'll remove this if we end up going with the approach that keeps `GetAssetInfoAsync`.

ahal:
Thanks for the great review! I've addressed the comments I could, and guessed what you'd prefer to do about `SearchAssetsAsync` not returning tag info. Lmk which option you prefer.

Personally I like this because calling `GetAssetInfoAsync` per asset is way more expensive than calling `SearchAssetsAsync` per tag.

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace ImmichFrame.Core.Logic.Pool;
+
+public class TagAssetsPool(IApiCache apiCache, ImmichApi immichApi, IAccountSettings accountSettings) : CachingApiAssetsPool(apiCache, immichApi, accountSettings)
+{
+    protected override async Task<IEnumerable<AssetResponseDto>> LoadAssets(CancellationToken ct = default)
+    {
+        var tagAssets = new List<AssetResponseDto>();
+        var allTags = await apiCache.GetOrAddAsync($"allTags_{accountSettings.ImmichServerUrl}",
+            () => immichApi.GetAllTagsAsync(ct));
+        var tagValueToTag = allTags.ToDictionary(t => t.Value);
```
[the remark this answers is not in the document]
ahal: Oops, I need to update the documentation.

I realized that in Immich tags are case sensitive (e.g `People` and `people` are two separate tags). So we need to be case sensitive too.

ImmichFrame.Core/Logic/Pool/TagAssetsPool.cs
```diff
@@ -0,0 +1,69 @@
...
+                var tagInfo = await immichApi.SearchAssetsAsync(metadataBody, ct);
+
+                itemsInPage = tagInfo.Assets.Items.Count;
+
+                // Attach the tag that matched this search
+                foreach (var asset in tagInfo.Assets.Items)
+                {
+                    if (seenIds.Contains(asset.Id))
+                    {
+                        continue;
```
JW-CH, line 59:
```suggestion
                        tagAssets.First(a => a.Id == asset.Id).Tags.Add(tag);
                        continue;
```

Something like this is needed to actually display all the tags on an item.

ImmichFrame.Core.Tests/Logic/Pool/TagAssetsPoolTests.cs
```diff
@@ -0,0 +1,329 @@
...
+        _mockImmichApi.Setup(api => api.SearchAssetsAsync(It.Is<MetadataSearchDto>(d => d.TagIds.Contains(tag1Id)), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(CreateSearchResult(tag1Assets, 2));
+        _mockImmichApi.Setup(api => api.SearchAssetsAsync(It.Is<MetadataSearchDto>(d => d.TagIds.Contains(tag2Id)), It.IsAny<CancellationToken>()))
+            .ReturnsAsync(CreateSearchResult(tag2Assets, 2));
+
+        // Act
+        var result = (await _tagAssetsPool.TestLoadAssets()).ToList();
+
+        // Assert - Should only include the shared asset once, plus the two unique assets
+        Assert.That(result.Count, Is.EqualTo(3));
```
JW-CH: Doing a check like this seems necessary to me: `Assert.That(result.First(x => x.Id == sharedAssetId).Tags.Count, Is.EqualTo(2));`

In general, I find the Tests quite big and confusing. Could you check them again? Sorry I am also not too familiar with unit tests, I hope this makes sense. The check above just validates, that an Image can have more than just one tag. (See my other review comment)

ahal:
> In general, I find the Tests quite big and confusing

Agreed, apologies for that. As you guessed I'm using Claude to help out. I think this next iteration is a lot more clear though. I also added an assertion that covers your first comment.

JW-CH:
@ahal Thank you for your contribution!

PR #549, opened 2026-01-07
Add setting to adjust temperature number display style

nopoz:
Adds boolean setting "UseWholeNumberTemperatures" to display temperature as integer (whole numbers) in order to simplify display. 

When set to false or not set, the default is to display Temperature as decimal. 

## Summary by CodeRabbit

- **New Features**
  - Added settings to show videos, display account tags, show tag descriptions, and play audio.
  - Added configurable weather temperature precision, supporting 0–2 decimal places and defaulting to 1.
  - Exposed the new settings through the client settings API.

- **Bug Fixes**
  - Weather temperatures now respect the configured decimal precision instead of always showing one decimal place.

- **Documentation**
  - Updated sample JSON, YAML, and environment configurations with the new options.

### DIFF
--- a/ImmichFrame.Core/Interfaces/IClientSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IClientSettings.cs
@@ -23,6 +23,7 @@ namespace ImmichFrame.Core.Interfaces
         public string Style { get; }
         public string? BaseFontSize { get; }
         public bool ShowWeatherDescription { get; }
+        public int TemperatureDecimalDigits { get; }
         public string? WeatherIconUrl { get; }
         public bool ImageZoom { get; }
         public bool ImagePan { get; }
--- a/ImmichFrame.WebApi.Tests/Models/TemperatureDecimalDigitsValidationTests.cs
+++ b/ImmichFrame.WebApi.Tests/Models/TemperatureDecimalDigitsValidationTests.cs
@@ -1,0 +1,50 @@
+using NUnit.Framework;
+using ImmichFrame.WebApi.Helpers;
+using ImmichFrame.WebApi.Models;
+
+namespace ImmichFrame.WebApi.Tests.Models;
+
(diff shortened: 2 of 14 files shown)


3rob3:
I would be good just always using a whole number. Feel like this has gone back and forth. This is fine too just would prefer less settings. 
If we do go with this you would have to add the tests also.

JW-CH:
Some people would like one digit, you like none, some like two or more... Did you check for a possibility to hand over a format instead of a bool? I'd prefer this more. I'm not sure how precise openweathermap delivers the data, might be worth checking.

3rob3:
I’d also like to remove the C/F and just use degree symbol.

nopoz:
> Some people would like one digit, you like none, some like two or more... Did you check for a possibility to hand over a format instead of a bool? I'd prefer this more. I'm not sure how precise openweathermap delivers the data, might be worth checking.

I assume you're playing devil's advocate and don't actually want 2 digits of temp displayed yourself? :sweat_smile:

Anyways, it's a fair point so I've changed the setting to `TemperatureDecimalDigits` which accepts a numerical value - the user can set to 0-2. The default is 1, the original one decimal place. 0 is none. 2 is for the special people. 

openweathermap API provides two decimal places in their temp values

nopoz:
> I’d also like to remove the C/F and just use degree symbol.

See https://github.com/immichFrame/ImmichFrame/pull/553

3rob3:
I'm good with this.

3rob3 (changes requested):
Can you update the Settings.example.json, Settings.example.yml, and example.env with this new setting please?

3rob3 (approved):
thanks!

nopoz:
@3rob3 @JW-CH rebased on current main. This had gone stale and was conflicting.

PR #553, opened 2026-01-07
Add ShowTemperatureUnit setting to hide F/C from weather display

nopoz:
- Add ShowTemperatureUnit boolean property (default: true) across backend and frontend
  - When false, displays only degree symbol (e.g., "72°") without F/C indicator
  - Handle undefined weather.unit with fallback to "°"
  - Update test data files (TestV2.json, TestV2.yml) for new setting

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration setting to control display of temperature units (enabled by default).
  * Clock/weather display now respects the temperature-unit setting.
  * Client and server configuration samples and environment vars updated to include the new setting.

### DIFF
--- a/docker/Settings.example.json
+++ b/docker/Settings.example.json
@@ -34,7 +34,8 @@
     "ImageZoom": true,
     "ImagePan": false,
     "ImageFill": false,
-    "Layout": "splitview"
+    "Layout": "splitview",
+    "ShowTemperatureUnit": true
   },
   "Accounts": [
     {
--- a/docker/Settings.example.yml
+++ b/docker/Settings.example.yml
@@ -33,6 +33,7 @@ General:
   ImagePan: false
   ImageFill: false
   Layout: splitview
+  ShowTemperatureUnit: true
 Accounts:
   - ImmichServerUrl: REQUIRED
     # Exactly one of ApiKey or ApiKeyFile must be set.
--- a/docker/example.env
+++ b/docker/example.env
@@ -46,3 +46,4 @@ ApiKey=KEY
 # WeatherLatLong=
 # Language=en
 # Webhook=
+# ShowTemperatureUnit=true


3rob3:
I'm good with this too.

JW-CH:
> I'm good with this too.

Sorry, I lack behind with review. Will probably find time this weekend.

ImmichFrame.WebApi/Helpers/Config/ServerSettingsV1.cs
```diff
@@ -120,6 +120,7 @@ class GeneralSettingsV1Adapter(ServerSettingsV1 _delegate) : IGeneralSettings
         public string Style => _delegate.Style;
         public string? BaseFontSize => _delegate.BaseFontSize;
         public bool ShowWeatherDescription => _delegate.ShowWeatherDescription;
+        public bool ShowTemperatureUnit => true;
```
JW-CH, line 123: why is this default true in the v1 settings?

immichFrame.Web/src/lib/components/elements/clock.svelte
```diff
@@ -80,7 +80,7 @@
 				{/if}
 				<div class="weather-location">{weather.location},</div>
 				<div class="weather-temperature">{weather.temperature?.toFixed(1)}</div>
-				<div class="weather-unit">{weather.unit}</div>
+				<div class="weather-unit">{$configStore.showTemperatureUnit === false ? '°' : (weather.unit ?? '°')}</div>
```
JW-CH, line 83: If `configStore.showTemperatureUnit` is false, you just show `°`? Shouldn't this just be empty?

JW-CH (changes requested):

immichFrame.Web/src/lib/components/elements/clock.svelte
3rob3: This is just to remove the C/F after degree symbol.

JW-CH: Oh, ok. Thank you.

3rob3:
Closing this, we removed temperature unit by default so this is no longer needed.

PR #561, opened 2026-01-16
Add "ExcludedPeople" setting to filter out specific people from "People" results

nopoz:
Rebased on current main.

`ExcludedPeople` is a per-account list of Immich person IDs. Any asset containing one of those people is dropped from the slideshow, the same way `ExcludedAlbums` drops an album's assets.

```yaml
Accounts:
  - ImmichServerUrl: http://your-server:2283
    ApiKey: "your-api-key"
    People:
      - 00000000-0000-0000-0000-000000000001  # show photos of this person
    ExcludedPeople:
      - 00000000-0000-0000-0000-000000000003  # but not if this person is in them too
```

One thing changed since the last review: exclusion now applies to every pool, not just `People`.

@3rob3 raised that this wouldn't work alongside Albums. That was true when I opened this PR, and it isn't anymore. `AlbumAssetsPool` no longer calls `GetAlbumInfoAsync`. Since a37014c and accad5a it goes through `/search/metadata` with `AlbumIds` and `WithPeople`, and `GetAlbumInfoAsync` is now unused anywhere in the codebase. The album/people asymmetry I described in January is gone.

So rather than special-casing albums, the filter now sits in `ApplyAccountFilters` next to the existing `ExcludedAlbums` line, which means the album, favorite, memory, people, tag and all-assets pools all get it. `PersonAssetsPool` is untouched by this PR now, where the earlier version rewrote it.

Measured against my own library, where the excluded person appears in 2481 assets:

| pool | before | after |
| --- | --- | --- |
| album | 3184 | 2839 |
| people | 1434 | 883 |
| favorites | 2 | 0 |
| memories | 7 | 6 |

For the all-assets pool a 400-asset control sample contained 44 of that person's assets, and none after filtering.

Excluded assets are matched by id rather than by reading `asset.People`. Reading `People` would avoid the extra fetch, but it depends on every pool setting `WithPeople`, and they don't: `GetExcludedAlbumAssets` doesn't set it, and memory assets arrive without it and only get it back-filled per asset. Matching by id costs one paged `personIds` search per excluded person, cached the same way the excluded-album lookup is, and keeps exclusion correct regardless of how a pool fetched its assets.

No Immich version change needed. The only call this adds is `/search/metadata` with `personIds`, which `PersonAssetsPool` already makes.

## Summary by CodeRabbit

* **New Features**
  * Added account-level configuration to exclude media featuring specific people.
  * Excluded people can be configured using person IDs in JSON, YAML, or environment settings.
  * Matching assets are filtered from displayed media, including across large collections and multiple people.

* **Documentation**
  * Updated sample configuration files with `ExcludedPeople` examples.

* **Tests**
  * Added coverage for pagination, multiple exclusions, empty settings, filtering, and request cancellation.


3rob3:
I'm surprised this was so simple. It has been awhile since I attempted this, but I thought there was an Immich API issue where something (albums maybe?) didn't include person info?

nopoz:
> I'm surprised this was so simple. It has been awhile since I attempted this, but I thought there was an Immich API issue where something (albums maybe?) didn't include person info?

You're right, there is a limitation with the Album results. 

  The Issue with Albums on the Immich side:

  1. `getAlbumInfo` has no `withPeople` parameter - it only accepts `id`, `key`, and `withoutAssets` (lines 940-966 in the OpenAPI spec)
  2. `people` is NOT a required field in `AssetResponseDto` - Looking at the OpenAPI spec (line 9299-9319), the required fields are: `checksum`, `deviceAssetId`, `deviceId`, `duration`, `fileCreatedAt`, etc. The `people` field (line 9245-9250) is optional.
  3. Other endpoints have `withPeople` - The `searchMetadata` and `searchRandom` endpoints have `WithPeople = true` parameters, which is why they can request people data. Album fetching doesn't have this option.

 The implication: When you call `GetAlbumInfoAsync`, the assets returned may have `People = null` because Immich doesn't guarantee that field is populated without explicit request.

To clarify the intent of this PR, the `ExcludedPeople` feature only filters out specific people from the `People` results because of the above limitations.

JW-CH:
Could you rebase this on master.

3rob3:
Could you rebase this on master?

nopoz:
@3rob3 @JW-CH rebased on current main, and this is a single feature again.

The Albums limitation I described in January doesn't apply anymore: `AlbumAssetsPool` stopped calling `GetAlbumInfoAsync` and now fetches through `/search/metadata` with `WithPeople`, so exclusion reaches album assets too. Excluding one person on my own library takes the album pool from 3184 assets to 2839. Numbers for the other pools are in the updated description.

3rob3:
Looks clean, and directly in line with ExcludedAlbums. I like it!

PR #569, opened 2026-01-19
Fix: Handle multiple weather icons

3rob3:
alternate implementation of #526 

## Release Notes

* **Bug Fixes**
  * Improved weather icon display handling to correctly extract and show the primary weather icon when multiple icons are available, ensuring icons only render when valid data is present.


JW-CH:
Closes #523

PR #595, opened 2026-02-19
Fix: Range header forwarding (Safari/iOS video fix)

3rob3:
## Summary by CodeRabbit

* **New Features**
  * Range-aware video playback endpoint with partial content (206), Accept-Ranges header and explicit 416 handling.
  * Client API: new range-aware playback method for streaming with range support.

* **Refactor**
  * Asset delivery now returns a structured AssetResponse including filename, content type, stream and metadata (content-range, content-length, partial flag, owner).

* **Bug Fixes**
  * Webhook notifications suppressed for range/partial requests to avoid redundant triggers.

### DIFF
--- a/ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
+++ b/ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
@@ -163,26 +163,49 @@ public class PooledImmichFrameLogic : IAccountImmichFrameLogic
     private async Task<(string fileName, string ContentType, Stream fileStream)> GetVideoAsset(Guid id)
     {
         var videoResponse = await _immichApi.PlayAssetVideoAsync(id, string.Empty);
-
         if (videoResponse == null)
             throw new AssetNotFoundException($"Video asset {id} was not found!");
 
-        var contentType = "";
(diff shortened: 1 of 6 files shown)


ImmichFrame.Core/Logic/PooledImmichFrameLogic.cs
```diff
@@ -160,29 +163,54 @@ public Task<IEnumerable<AssetResponseDto>> GetAssets()
         return (fileName, contentType, data.Stream);
     }
 
-    private async Task<(string fileName, string ContentType, Stream fileStream)> GetVideoAsset(Guid id)
+    private async Task<FileResponse> PlayVideoWithRange(Guid id, string rangeHeader, CancellationToken cancellationToken = default)
     {
-        var videoResponse = await _immichApi.PlayAssetVideoAsync(id, string.Empty);
+        var url = $"{_immichApi.BaseUrl.TrimEnd('/')}/assets/{id}/video/playback";
```
JW-CH: I don't like this 'hard coded' URL this much. Could you either find a way to use a class generated in immichapi or at least add a method to the partial implementation if immichapi?

PR #610, opened 2026-03-23
feat: add generic custom data source widget

RC1140:
## Summary

Adds a configurable widget that fetches structured data from user-specified URLs and displays it on the photo frame. This enables displaying arbitrary external data (fitness stats, Home Assistant sensors, weather from custom sources, etc.) alongside photos.

- Supports multiple data sources, each with independent refresh intervals and caching
- URLs stay server-side (proxied via `/api/CustomWidget`) — only display config is exposed to the client
- Follows existing widget patterns (weather/calendar/appointments) for consistency

## Configuration

```yaml
General:
  ShowCustomWidget: true
  CustomWidgetPosition: 'top-left'  # top-left, top-right, bottom-left, bottom-right
  CustomWidgetSources:
    - Url: 'http://my-service:3001/api/stats'
      RefreshMinutes: 10
    - Url: 'http://another-service/data'
      RefreshMinutes: 30
```

## Expected response format from data sources

```json
{
  "title": "My Stats",
  "items": [
    { "label": "Metric A", "value": "42", "secondary": "optional detail" },
    { "label": "Metric B", "value": "100" }
  ]
}
```

## Changes

**Backend (C# / ASP.NET Core):**
- `CustomWidgetSourceConfig` model for per-source URL + refresh interval
- `CustomWidgetData` / `CustomWidgetItem` response models
- `ICustomWidgetService` + `CustomWidgetService` with per-source caching
- `CustomWidgetController` — proxies data (keeps URLs server-side)
- Settings: `ShowCustomWidget`, `CustomWidgetSources`, `CustomWidgetPosition`
- `ClientSettingsDto` exposes display config only

**Frontend (Svelte 5):**
- New `custom-widget.svelte` component with configurable corner positioning
- Integrated into `home-page.svelte` with conditional rendering
- Regenerated oazapfts API client with new endpoint types

## Test plan

- [ ] Verify widget renders with valid data source URL
- [ ] Verify widget hidden when `ShowCustomWidget: false`
- [ ] Verify multiple data sources display correctly
- [ ] Verify positioning works for all four corners
- [ ] Verify graceful handling when data source is unreachable

## Summary by CodeRabbit

* **New Features**
  * Custom widget component now available on the home page for displaying external data from configured sources
  * Users can configure widget visibility, position, data source URLs, and refresh intervals
  * Automatic data refresh based on configured intervals per source

### DIFF
--- a/ImmichFrame.Core/Interfaces/ICustomWidgetService.cs
+++ b/ImmichFrame.Core/Interfaces/ICustomWidgetService.cs
@@ -1,0 +1,6 @@
+using ImmichFrame.Core.Models;
+
+public interface ICustomWidgetService
+{
+    Task<List<CustomWidgetData>> GetCustomWidgetData();
+}
--- a/ImmichFrame.Core/Services/CustomWidgetService.cs
+++ b/ImmichFrame.Core/Services/CustomWidgetService.cs
@@ -1,0 +1,74 @@
+using System.Net.Http.Json;
+using ImmichFrame.Core.Interfaces;
+using ImmichFrame.Core.Models;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+
+public class CustomWidgetService : ICustomWidgetService
+{
+    private readonly IGeneralSettings _settings;
+    private readonly ILogger<CustomWidgetService> _logger;
+    private readonly IHttpClientFactory _httpClientFactory;
+    private readonly IMemoryCache _cache = new MemoryCache(new MemoryCacheOptions());
+
+    public CustomWidgetService(IGeneralSettings settings, ILogger<CustomWidgetService> logger, IHttpClientFactory httpClientFactory)
+    {
+        _settings = settings;
+        _logger = logger;
+        _httpClientFactory = httpClientFactory;
+    }
+
+    public async Task<List<CustomWidgetData>> GetCustomWidgetData()
+    {
+        var results = new List<CustomWidgetData>();
+
+        foreach (var source in _settings.CustomWidgetSources)
+        {
+            if (string.IsNullOrWhiteSpace(source.Url))
+                continue;
+
+            try
+            {
+                var cacheKey = $"customwidget_{source.Url}";
+                var data = await _cache.GetOrCreateAsync(cacheKey, async entry =>
+                {
+                    entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromMinutes(source.RefreshMinutes);
+                    return await FetchWidgetData(source.Url);
+                });
+
+                if (data != null)
+                    results.Add(data);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Failed to fetch custom widget data from '{Url}'", source.Url);
+            }
+        }
+
+        return results;
+    }
+
+    private async Task<CustomWidgetData?> FetchWidgetData(string url)
+    {
+        var client = _httpClientFactory.CreateClient();
+
+        try
+        {
+            var response = await client.GetAsync(url);
+
+            if (!response.IsSuccessStatusCode)
+            {
+                _logger.LogError("Custom widget source '{Url}' returned status {StatusCode}", url, response.StatusCode);
+                return null;
+            }
+
+            var data = await response.Content.ReadFromJsonAsync<CustomWidgetData>();
+            return data;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error fetching custom widget data from '{Url}'", url);
+            return null;
+        }
+    }
+}
--- a/ImmichFrame.WebApi/Controllers/CustomWidgetController.cs
+++ b/ImmichFrame.WebApi/Controllers/CustomWidgetController.cs
@@ -1,0 +1,30 @@
+using ImmichFrame.Core.Interfaces;
+using ImmichFrame.Core.Models;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace ImmichFrame.WebApi.Controllers
+{
+    [ApiController]
+    [Route("api/[controller]")]
+    [Authorize]
+    public class CustomWidgetController : ControllerBase
+    {
+        private readonly ILogger<AssetController> _logger;
+        private readonly ICustomWidgetService _customWidgetService;
+
+        public CustomWidgetController(ILogger<AssetController> logger, ICustomWidgetService customWidgetService)
+        {
+            _logger = logger;
+            _customWidgetService = customWidgetService;
+        }
+
+        [HttpGet(Name = "GetCustomWidgetData")]
+        public async Task<List<CustomWidgetData>> GetCustomWidgetData(string clientIdentifier = "")
+        {
+            var sanitizedClientIdentifier = clientIdentifier.SanitizeString();
+            _logger.LogDebug("Custom widget data requested by '{sanitizedClientIdentifier}'", sanitizedClientIdentifier);
+            return await _customWidgetService.GetCustomWidgetData();
+        }
+    }
+}
--- a/ImmichFrame.Core/Interfaces/IServerSettings.cs
+++ b/ImmichFrame.Core/Interfaces/IServerSettings.cs
@@ -1,4 +1,6 @@
-﻿namespace ImmichFrame.Core.Interfaces
+﻿using ImmichFrame.Core.Models;
+
+namespace ImmichFrame.Core.Interfaces
 {
     public interface IServerSettings
     {
@@ -66,6 +68,9 @@
         public bool PlayAudio { get; }
         public string Layout { get; }
         public string Language { get; }
+        public bool ShowCustomWidget { get; }
+        public List<CustomWidgetSourceConfig> CustomWidgetSources { get; }
+        public string CustomWidgetPosition { get; }
 
         public void Validate();
     }
--- a/ImmichFrame.Core/Models/CustomWidgetData.cs
+++ b/ImmichFrame.Core/Models/CustomWidgetData.cs
@@ -1,0 +1,14 @@
+namespace ImmichFrame.Core.Models;
+
+public class CustomWidgetData
+{
+    public string Title { get; set; } = string.Empty;
+    public List<CustomWidgetItem> Items { get; set; } = new();
+}
+
+public class CustomWidgetItem
+{
+    public string Label { get; set; } = string.Empty;
+    public string Value { get; set; } = string.Empty;
+    public string? Secondary { get; set; }
+}
--- a/ImmichFrame.Core/Models/CustomWidgetSourceConfig.cs
+++ b/ImmichFrame.Core/Models/CustomWidgetSourceConfig.cs
@@ -1,0 +1,7 @@
+namespace ImmichFrame.Core.Models;
+
+public class CustomWidgetSourceConfig
+{
+    public string Url { get; set; } = string.Empty;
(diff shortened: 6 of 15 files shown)


JW-CH:
Do you have any example sources by chance to test?

RC1140:
@JW-CH Sorry I built that out myself will add some screenshots in a bit.

This is what it looks like when rendered using the webview 
image

As mentioned the API endpoint I hit to get my strength data is a custom service and not really anything public, but returning the data in the structure below works for me atleast 

```
 {
    "title": "IRON 3",
    "items": [
      { "label": "Last Workout", "value": "Full Body + Accessory", "secondary": "3h ago" },
      { "label": "Squat", "value": "30 kg x 5x5" },
      { "label": "Bench", "value": "10 kg x 5x5" },
      { "label": "Deadlift", "value": "30 kg x 1x5" },
      { "label": "OHP", "value": "10 kg x 5x5" },
      { "label": "Row", "value": "15 kg x 5x5" },
      { "label": "Week Volume", "value": "1,735 kg" },
      { "label": "Streak", "value": "4 weeks" },
      { "label": "Fitness (CTL)", "value": "21" },
      { "label": "Fatigue (ATL)", "value": "9" },
      { "label": "Resting HR", "value": "51 bpm" },
      { "label": "HRV", "value": "44" },
      { "label": "Rides", "value": "1 (60min)" }
    ]
  }
````

The idea here is that you have a title and then any number of rows returned which then gets rendered. You can add multiple sources if your data is not centralized

**Comment by RC1140** [comment]
Dont know how much extra work this would be but you can test it with a small python server like this 

```
python3 -c "
import json
from http.server import HTTPServer, BaseHTTPRequestHandler

data = {'title':'IRON 3','items':[{'label':'Last Workout','value':'Full Body + Accessory','secondary':'3h ago'},{'label':'Squat','value':'30 kg x 5x5'},{'label':'Bench','value':'10 kg x 5x5'},{'label':'Deadlift','value':'30 kg x 1x5'},{'label':'OHP','value':'10 kg x 5x5'},{'label':'Row','value':'15 kg x 5x5'},{'label':'Week Volume','value':'1,735 kg'},{'label':'Streak','value':'4 weeks'},{'label':'Fitness (CTL)','value':'21'},{'label':'Fatigue (ATL)','value':'9'},{'label':'Resting HR','value':'51 bpm'},{'label':'HRV','value':'44'},{'label':'Rides','value':'1 (60min)'}]}

class H(BaseHTTPRequestHandler):
    def do_GET(self):
        body = json.dumps(data).encode()
        self.send_response(200)
        self.send_header('Content-Type', 'application/json')
        self.send_header('Content-Length', len(body))
        self.end_headers()
        self.wfile.write(body)
    def log_message(self, *a): pass

HTTPServer(('', 8080), H).serve_forever()
"
```


Paste that in a console and hit localhost:8080

File: ImmichFrame.WebApi/Controllers/CustomWidgetController.cs [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+{
+    [ApiController]
+    [Route("api/[controller]")]
+    [Authorize]
+    public class CustomWidgetController : ControllerBase
+    {
+        private readonly ILogger<AssetController> _logger;
+        private readonly ICustomWidgetService _customWidgetService;
+
+        public CustomWidgetController(ILogger<AssetController> logger, ICustomWidgetService customWidgetService)
```
JW-CH, line 16: Wrong Logger-Type (AssetController)

ImmichFrame.Core/Interfaces/ICustomWidgetService.cs
```diff
@@ -0,0 +1,6 @@
+using ImmichFrame.Core.Models;
+
```
JW-CH, line 2: Namespace is missing

ImmichFrame.Core/Services/CustomWidgetService.cs
```diff
@@ -0,0 +1,74 @@
+using System.Net.Http.Json;
+using ImmichFrame.Core.Interfaces;
+using ImmichFrame.Core.Models;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+
```
JW-CH, line 6: Namespace is missing

ImmichFrame.WebApi/Controllers/CustomWidgetController.cs
```diff
@@ -0,0 +1,30 @@
...
+using ImmichFrame.Core.Models;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace ImmichFrame.WebApi.Controllers
+{
+    [ApiController]
+    [Route("api/[controller]")]
+    [Authorize]
+    public class CustomWidgetController : ControllerBase
```
JW-CH, line 11: I'd rename the controller to something like 'WidgetController' - Maybe in the Future there could be more Widgets maybe?

ImmichFrame.WebApi/Controllers/CustomWidgetController.cs
```diff
@@ -0,0 +1,30 @@
...
+        private readonly ICustomWidgetService _customWidgetService;
+
+        public CustomWidgetController(ILogger<AssetController> logger, ICustomWidgetService customWidgetService)
+        {
+            _logger = logger;
+            _customWidgetService = customWidgetService;
+        }
+
+        [HttpGet(Name = "GetCustomWidgetData")]
+        public async Task<List<CustomWidgetData>> GetCustomWidgetData(string clientIdentifier = "")
```
JW-CH, line 23: Naming here is ok

ImmichFrame.Core/Services/CustomWidgetService.cs
```diff
@@ -0,0 +1,74 @@
...
+public class CustomWidgetService : ICustomWidgetService
+{
+    private readonly IGeneralSettings _settings;
+    private readonly ILogger<CustomWidgetService> _logger;
+    private readonly IHttpClientFactory _httpClientFactory;
+    private readonly IMemoryCache _cache = new MemoryCache(new MemoryCacheOptions());
```
JW-CH, line 12: Could you use the `ApiCache` here, see `OpenWeatherMapService` as reference

JW-CH (changes requested):
Hey, thank you very much for your contribution! Really good implementation following the standards of the existing services.

I did an initial review with some first comments, I will also add another comment on the 'structure' of the expected Api-Data.

JW-CH:
I did a quick Claude research - posting the response after. I quite like the approach claude suggests and checked if this would be possible. Let me know what you think:

Clause answer:
===============

The current design requires every external API to already return data in the `CustomWidgetData` shape. This means Home Assistant, OpenHAB, Grafana, or any other common self-hosted service would need a proxy/transformer in between — which adds friction for users.
                                                                                                                                                                                                                                                                                                                                                                                                                                                    
A more open approach would be to add JSONPath mapping + header support directly to `CustomWidgetSourceConfig`, so ImmichFrame can talk to arbitrary REST APIs natively:                                                                                                                                                                                                                                                                             
 ```                                                                                                                                                                                                                                                                                                                                                                                                                                                   
  public class CustomWidgetSourceConfig                                                                                                                                                                                                                                                                                                                                                                                                             
  {               
      public string Url { get; set; } = string.Empty;                                                                                                                                                                                                                                                                                                                                                                                               
      public int RefreshMinutes { get; set; } = 10;
      public Dictionary<string, string>? Headers { get; set; }  // e.g. Authorization: Bearer <token>                                                                                                                                                                                                                                                                                                                                               
      public string? Title { get; set; }                         // static widget title                                                                                                                                                                                                                                                                                                                                                             
      public string? ItemsPath { get; set; }                     // JSONPath to the items array, e.g. "$"
      public string? LabelPath { get; set; }                     // e.g. "$.attributes.friendly_name"                                                                                                                                                                                                                                                                                                                                               
      public string? ValuePath { get; set; }                     // e.g. "$.state"
      public string? SecondaryPath { get; set; }                 // e.g. "$.attributes.unit_of_measurement"                                                                                                                                                                                                                                                                                                                                         
  }               
```
A Home Assistant config would then look like this — no proxy needed:                                                                                                                                                                                                                                                                                                                                                                              
```
  {                                                                                                                                                                                                                                                                                                                                                                                                                                                 
    "Url": "http://homeassistant:8123/api/states?domain=sensor",
    "Headers": { "Authorization": "Bearer <token>" },
    "Title": "Sensors",                                                                                                                                                                                                                                                                                                                                                                                                                             
    "ItemsPath": "$",
    "LabelPath": "$.attributes.friendly_name",                                                                                                                                                                                                                                                                                                                                                                                                      
    "ValuePath": "$.state",
    "SecondaryPath": "$.attributes.unit_of_measurement"
  }                                                                                                                                                                                                                                                                                                                                                                                                                                                 
```
  If none of the path fields are set, the service falls back to deserializing the response as `CustomWidgetData` directly — so the existing generic use case stays fully supported and backwards-compatible.                                                                                                                                                                                                                                          
                  
  The main trade-off is a JSONPath dependency (e.g. `JsonPath.Net`), but it's a small, well-maintained library and the flexibility gain is significant for the self-hosted audience this project targets.

**Comment by JW-CH** [comment]
One thing the JSONPath mapping doesn't yet solve: selective filtering of items from an array response.                                                                                                                                                                                                                                                                                                                                            

For example, querying Home Assistant's `/api/states?domain=sensor` returns every sensor. There's no way to say "only show these three." You'd have to configure one source entry per entity, which gets verbose quickly.

A small addition to `CustomWidgetSourceConfig` would fix this:                                                                                                                                                                                                                                                                                                                                                                                      

```cs
public List<string>? Include { get; set; }  // only render items whose resolved label matches
```

Config example:
```json
  {
    "Url": "http://homeassistant:8123/api/states?domain=sensor",
    "Headers": { "Authorization": "Bearer <token>" },                                                                                                                                                                                                                                                                                                                                                                                               
    "Title": "Sensors",
    "ItemsPath": "$",                                                                                                                                                                                                                                                                                                                                                                                                                               
    "LabelPath": "$.attributes.friendly_name",
    "ValuePath": "$.state",
    "SecondaryPath": "$.attributes.unit_of_measurement",                                                                                                                                                                                                                                                                                                                                                                                            
    "Include": ["Living Room Temp", "Outdoor Humidity"]
  }
```
The service would filter the resolved items against the `Include` list after applying the JSONPath mappings. If `Include` is `null` or empty, all items are shown.                                                                                                                                                                                                                                                         
   
  ---                                                                                                                                                                                                                                                                                                                                                                                                                                               
  Single-entity endpoints are also worth considering. HA exposes individual entities directly:
                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  `GET /api/states/sensor.ikea_inspelning_livingroom_power`
                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  This returns a single object, not an array, so `ItemsPath` wouldn't apply. A clean way to handle this: if `ItemsPath` is not set, treat the entire response as a single item rather than trying to iterate over it.                                                                                                                                                                                                                                  
```json
{                                                                                                                                                                                                                                                                                                                                                                                                                                                 
    "Url": "http://homeassistant:8123/api/states/sensor.ikea_inspelning_livingroom_power",
    "Headers": { "Authorization": "Bearer <token>" },
    "Title": "Office",                                                                                                                                                                                                                                                                                                                                                                                                                              
    "LabelPath": "$.attributes.friendly_name",
    "ValuePath": "$.state",                                                                                                                                                                                                                                                                                                                                                                                                                         
    "SecondaryPath": "$.attributes.unit_of_measurement"
  }
```
This keeps the config intuitive: `ItemsPath` present → array response, `ItemsPath` absent → single object response.

Issue #464, opened 2025-09-17
Splitview causes the same photo to show twice sometimes


Chaoscontrol:
### Version

1.0.28.0

### Deployment Method

Desktop and Android

### Description

I have splitview enabled by default. It's an album with not too many photos. 

Sometimes when a splitview image shows, it shows the same photo twice, side by side. I guess it randomly finds another portrait photo to show, and since there's not a lot, eventually it finds the exact same by chance.

Logs are probably not relevant as I'm running it as a Home Assistant addon. 

### Reproduction

1. Enable splitview. 
2. Filter with an album with a low amount of pictures (still enough portrait ones so it doesn't ALWAYS happen). 
3. Wait until eventually a portrait photo will show twice. 

### Expectations

Same photo to be excluded when finding the second picture.

### Configuration

```xml
General:
  ClockFormat: "HH:mm"
  # WeatherApiKey: redacted
  # UnitSystem: metric
  # WeatherLatLong: 53.380714,-2.1678159999999997
  ShowPhotoDate: false
  ShowAlbumName: false
  ShowImageLocation: false
  ShowPeopleDesc: false
  ShowImageDesc: false
  ShowWeatherDescription: false
  ShowProgressBar: false
  PrimaryColor: "#fff"

  Interval: 300
  ImageFill: true


Accounts:
  # - ImmichServerUrl: "http://192.168.1.158:8181"
  #   ApiKey: "redacted"
  #   ImagesFromDays: 60
  #   Albums: []
  #   ExcludedAlbums: []
  #   People:
  #     - "79f8359a-bd6b-4fc1-ad71-7e4ad1aa032a"
  - ImmichServerUrl: "http://192.168.1.158:8181"
    ApiKey: "redacted"
    # ImagesFromDays: 60
    Albums:
      - "4c7d865a-1ce7-49bd-884f-8541c9d9a778"
    ExcludedAlbums: []
    # People:
    #   - "0bf41189-7e74-4867-b94a-497c84806c6b"
```

### Logs

```shell
Starting...
/etc/cont-init.d/00-banner.sh: executing
-----------------------------------------------------------
 Add-on: Immich Frame
 Display your immich gallery as a digital photo frame
-----------------------------------------------------------
 Add-on version: 1.0.28.0
 You are running the latest version of this add-on.
 System: Home Assistant OS 16.2
 Architecture: amd64 / generic-x86-64
 Home Assistant Core: 2025.9.3
[...]

PR #612, opened 2026-03-25
avoid duplicates in split view

Qoen1:
# summary
On my immich frame instance I mainly want to show memories which there are not a whole lot of and I noticed that the side by side view sometimes shows duplicates. I don't think it should be that way, so i've implemented a small client-side function that will remove the duplicate picture based on it's id and show just a single photo.

# what did I change?
Since I don't know this codebase well and this is my first time contributing to an open-source project I thought i'd keep my change as small as possible so I just added a function to the home page. Let me know if this should be implemented in some other way!

# side effects
if you have only one image in the queue the ui looks the same regardless of the chosen layout in the config.

# issues
I saw someone mention this in: https://github.com/immichFrame/ImmichFrame/issues/464

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate asset display when loading content in split-view and standard layouts so upcoming assets no longer repeat.
  * Improved browsing consistency by filtering the next-assets list and, when possible, substituting repeated items with a non-duplicate portrait from the backlog to keep displayed sets unique.

### DIFF
--- a/immichFrame.Web/src/lib/components/home-page/home-page.svelte
+++ b/immichFrame.Web/src/lib/components/home-page/home-page.svelte
@@ -200,13 +200,7 @@
 	}
 
 	function removeDuplicateFromNextAssets(next: api.AssetResponseDto[]): api.AssetResponseDto[]{
-		// while(isDuplicate(next) && assetBacklog.length > 0){
-		// 	const duplicatePicture = next[0];
-		// 	next = assetBacklog.splice(0,1)
-		// 	next.push(duplicatePicture);
-		// }
 		if(isDuplicate(next)){
-			console.log("Duplicate detected!")
 			return next.slice(0,1);
 		}
 		return next;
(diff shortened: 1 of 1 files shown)


JW-CH:
Hm, I think I'd prefer having this not client-side. The change itself looks good, but having the backend doing the work would simplify this I think. What are your thoughts on this?

Qoen1:
I thought that aswell, but it appears to me that the backend makes a list of all assets in the selection, which can contain the same asset multiple times, and sends it to the client. It does not care how the client displays the data. The client then receives that info and continuously picks two off of the list to display.

because this logic is already handled client-side i thought it would be more logical to put this logic there aswell

Qoen1:
hi @JW-CH I was thinking some more on your comment and I think you are right that it could be better to implement this server-side. I saw some othe pull requests that do something with ordering the queue of pictures, and I had nothing better to do so I implemented it in a new PR in a way I think can be easily expanded to include more features! Could you take a looka t it and let me know what you think? https://github.com/immichFrame/ImmichFrame/pull/616

PR #616, opened 2026-03-31
avoid duplicates in split view

Qoen1:
server-side alternative to: https://github.com/immichFrame/ImmichFrame/pull/612

# summary
On my immich frame instance I mainly want to show memories which there are not a whole lot of and I noticed that the side by side view sometimes shows duplicates. I don't think it should be that way, so i've implemented a server-side chain  class that can manipulate queues (in this case for removing duplicates) and can be expanded to fit other que manipulation usecases.

# side effects
the amount of times only one portrait image is show was higher during testing because duplicates are moved to the back of the queue. I think this can best be fixed by adding another implementation of QueueManipulator that tries to optimize for this but perhaps it is best to leave that for another time.

# issues
I saw someone mention this in: https://github.com/immichFrame/ImmichFrame/issues/464

## Summary by CodeRabbit

* **New Features**
  * Queue-based asset processing with randomized ordering and avoidance of consecutive portrait-image duplicates
  * Image orientation detection to classify assets as landscape or portrait from EXIF metadata

* **Tests**
  * Added tests validating mutation behavior for empty inputs, non-duplicates, separated/adjacent duplicates, batch boundaries, and orientation-sensitive cases

### DIFF
--- a/ImmichFrame.Core/Helpers/ImageHelper.cs
+++ b/ImmichFrame.Core/Helpers/ImageHelper.cs
@@ -18,9 +18,9 @@ namespace ImmichFrame.Core.Helpers
         public static double GetAspectRatioFloat(double width, double height)
         {
             if (width <= 0 || height <= 0)
-                throw new ArgumentException("Width and height must be positive integers.");
+                throw new ArgumentException("Width and height must be positive.");
         
-            return (double)width / height;
+            return width / height;
         }
 
         public static bool IsLandscape(double exifWidth, double exifHeight, string exifOrientation)
--- a/ImmichFrame.Core/Helpers/ImageHelper.cs
+++ b/ImmichFrame.Core/Helpers/ImageHelper.cs
@@ -14,5 +14,40 @@ namespace ImmichFrame.Core.Helpers
 
             return new MemoryStream(binData);
         }
+        
+        public static double GetAspectRatioFloat(double width, double height)
+        {
+            if (width <= 0 || height <= 0)
+                throw new ArgumentException("Width and height must be positive integers.");
+        
+            return (double)width / height;
+        }
+
+        public static bool IsLandscape(double exifWidth, double exifHeight, string exifOrientation)
+        {
+            double width;
+            double height;
+            if (exifOrientation == "1" 
+                || exifOrientation == "2"
+                || exifOrientation == "3"
+                || exifOrientation == "4")
+            {
+                width = exifWidth;
+                height = exifHeight;
+            }else if (exifOrientation == "5" 
+                      || exifOrientation == "6"
+                      || exifOrientation == "7"
+                      || exifOrientation == "8")
+            {
+                height = exifWidth;
+                width = exifHeight;
+            }
+            else
+            {
+                throw new ArgumentException(nameof(exifOrientation));
+            }
+
+            return GetAspectRatioFloat(width, height) > 1;
+        }
     }
 }
--- a/ImmichFrame.Core/Logic/QueueMutator/DuplicateAvoidingQueueMutator.cs
+++ b/ImmichFrame.Core/Logic/QueueMutator/DuplicateAvoidingQueueMutator.cs
@@ -1,0 +1,69 @@
+using ImmichFrame.Core.Api;
+using ImmichFrame.Core.Helpers;
+using Microsoft.Extensions.Logging;
+
+namespace ImmichFrame.Core.Logic.QueueMutator;
+
+public class DuplicateAvoidingQueueMutator: QueueMutatorBase<AssetResponseDto>
+{
+    private readonly int _batchSize;
+    private readonly ILogger<DuplicateAvoidingQueueMutator> _logger;
+
+    public DuplicateAvoidingQueueMutator(int batchSize, ILogger<DuplicateAvoidingQueueMutator> logger)
+    {
+        if(batchSize <= 0) throw new ArgumentException("Batch size must be greater than 0", nameof(batchSize));
+        _batchSize = batchSize;
+        _logger = logger;
+    }
+    public override IEnumerable<AssetResponseDto> Mutate(IEnumerable<AssetResponseDto> source)
+    {
+        var remaining = new Queue<AssetResponseDto>(source);
+        List<AssetResponseDto> result = new(remaining.Count);
+        var batch = new List<AssetResponseDto>(_batchSize);
+
+        int consecutiveDuplicates = 0;
+        while (remaining.Count > 0)
+        {
+            var currentItem = remaining.Dequeue();
+
+            bool landscape = false;
+            try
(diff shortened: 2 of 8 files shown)


ImmichFrame.WebApi/Program.cs
```diff
@@ -73,6 +76,13 @@ _ _ __ ___  _ __ ___  _  ___| |__ | |_ _ __ __ _ _ __ ___   ___
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen();
 
+builder.Services.AddTransient<IQueueMutator<AssetResponseDto>>(x =>
+{
+    var mutator = new RandomQueueMutator<AssetResponseDto>();
+    mutator.SetNext(new DuplicateAvoidingQueueMutator(2, x.GetService<ILogger<DuplicateAvoidingQueueMutator>>()!));
+    return mutator;
+});
```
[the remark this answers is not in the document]
Qoen1: I am not sure where the mutator will be used in the future so I prefer transient for now.

JW-CH:
I like the dutch words in the tests 😆 

Why are landscape images excluded from duplicate detection? I think a general deduplication in a batch would be nice (landscape and portrait). Can you elaborate a bit about your pattern choice. Why did you chose this 'Mutator'-Pattern? What code did you write yourself, what was AI-assisted?

Qoen1:
@JW-CH I excluded landscape images from duplicate detection because they are not displayed side-by-side (or at least I think so). So it does not matter if two are next to eachother. But I can change it to account for more images if that's better.

'Mutator' is just a term I thought sounded nice and accurately describes what they are supposed to do, so not something deliberately according to a design pattern. The design pattern I used is a [chain of responsibility ](https://refactoring.guru/design-patterns/chain-of-responsibility)because I think it fits very nicely. This way, in order to apply new features in ordereing the image queue, you can just implement another QueueMutator. But maybe I should be more clear about the patters I used? 

This pull request was made with 100% human-made fair trade code :)

Qoen1:
Oh and what do you think of the discussion about adding the ability to enable or diable mutators in the config file, and adding more robust pairing to the api so that the paired photos will not change in some edge cases as mentioned in https://github.com/immichFrame/ImmichFrame/pull/568#issuecomment-4211389563 ?

Qoen1:
@JW-CH I am not sure how to continue with this feature. Could you please have a look at my questions in this thread?

JW-CH:
Hey @Qoen1, sorry for the late reply, I was on vacation :P I want to work on the open PRs in the next days, thank you for your time.

Qoen1:
Hey @JW-CH I don't want to be too pushy about this but it would be a shame it this feature would be forgotten, so if you find the time at some point to take another look that would be great!

JW-CH:
@Qoen1 Sorry, again! I am digging into it, real cool stuff! I had to do some research on the 'chain of responsibility'-Pattern and I agree, this fits well. I'll spend a bit more time on reviewing the actual 'deduplication' code and need to test if this works with the logic that happens in the frontend, also how this would be extendible if the device could detect if it is in landscape or portrait mode. Nice stuff.

JW-CH:
The server uses fixed batches of 2 from index 0 (landscape also fills a slot). Example after shuffle (P=portrait, L=landscape):
  index:  0    1    2
  asset:  L1   P2   P2
- Server: batches [L1,P2] and [P2,…] → two P2 in different batches
- Client: [L1,P2] isn't two portraits → shows L1 alone → next head [P2,P2] → split renders P2 next to P2 = the duplicate.

ImmichFrame.Core/Helpers/ImageHelper.cs
```diff
@@ -14,5 +14,40 @@ public static Stream SaveDataUrlToStream(string dataUrl)
...
+            {
+                height = exifWidth;
+                width = exifHeight;
+            }
+            else
+            {
+                throw new ArgumentException(nameof(exifOrientation));
+            }
+
+            return GetAspectRatioFloat(width, height) > 1;
```
JW-CH, line 50: The client gates split view on `isPortrait`, but the server dedupes on `!IsLandscape()`. These match for normal images but diverge on square ones: `isPortrait` returns false (no split), while `!IsLandscape()` returns `true` (deduped). Could you switch the server to a portrait check to streamline both implementations?

ImmichFrame.Core/Logic/QueueMutator/DuplicateAvoidingQueueMutator.cs
```diff
@@ -0,0 +1,69 @@
...
+    private readonly int _batchSize;
+    private readonly ILogger<DuplicateAvoidingQueueMutator> _logger;
+
+    public DuplicateAvoidingQueueMutator(int batchSize, ILogger<DuplicateAvoidingQueueMutator> logger)
+    {
+        if(batchSize <= 0) throw new ArgumentException("Batch size must be greater than 0", nameof(batchSize));
+        _batchSize = batchSize;
+        _logger = logger;
+    }
+    public override IEnumerable<AssetResponseDto> Mutate(IEnumerable<AssetResponseDto> source)
```
JW-CH, line 18: The server-side logic **ignores videos** and also does fixed batches of two, this has to be changed to match the client-logic. (See my other comment)

JW-CH (changes requested):

ImmichFrame.Core/Logic/QueueMutator/DuplicateAvoidingQueueMutator.cs
Zaida-3dO: I feel trying to keep the server-side and client side logic in sync like this is fragile as it could accidentally diverge.

Zaida-3dO: Can we instead have this PR just prevent duplicates in a single asset batch.
This indirectly accomplishes the same thing. No duplicates in a batch means no duplicates side by side.

That way we don't need all the isPortrait/isLandscape logic which can go out of sync with the client

Zaida-3dO:
> the amount of times only one portrait image is show was higher during testing because duplicates are moved to the back of the queue.

Hey, @Qoen1 chiming in here from https://github.com/immichFrame/ImmichFrame/pull/568

My PR should handle sorting this out on the client side

ImmichFrame.Core/Logic/QueueMutator/DuplicateAvoidingQueueMutator.cs
Zaida-3dO: But this would mean that the batch sizes are non deterministic, not sure if that could cause any issues. I don't think it will.

what do you think of this @JW-CH updating this to a DuplicateRemovingQueueMutator

PR #628, opened 2026-04-11
Show shared albums

### LINKED ISSUE #625: Ability to show all albums
### Description

I created separate user in Immich which is connected with the frame. I share albums to this user and I want to configure ImmichFrame to show all of these shared albums instead of providing specific ids.

### Use Case

In this specific case, frame user has empty timeline and only shared albums. ImmichFrame doesn't show these albums.

### Details

_No response_

### Pre-Submission Checklist

3rob3:
closes #625 

## Summary by CodeRabbit

* **Chores**
  * Album-related API requests now automatically include the required query parameter for shared albums when it’s not already present.
  * Detection handles existing query strings and recognizes the parameter regardless of case, preserving other query parameters and separators.

### DIFF
--- a/ImmichFrame.Core/Api/ImmichApi.cs
+++ b/ImmichFrame.Core/Api/ImmichApi.cs
@@ -6,8 +6,7 @@
         {
             BaseUrl = url + "/api";
             _httpClient = httpClient;
-            _settings = new Lazy<Newtonsoft.Json.JsonSerializerSettings>(CreateSerializerSettings);
-            
+            _settings = new Lazy<Newtonsoft.Json.JsonSerializerSettings>(CreateSerializerSettings);            
         }
 
         public async Task<FileResponse> PlayAssetVideoWithRangeAsync(Guid id, string rangeHeader, CancellationToken cancellationToken = default)
--- a/ImmichFrame.Core/Api/ImmichApi.cs
+++ b/ImmichFrame.Core/Api/ImmichApi.cs
@@ -6,7 +6,7 @@
         {
             BaseUrl = url + "/api";
             _httpClient = httpClient;
-            _settings = new Lazy<Newtonsoft.Json.JsonSerializerSettings>(CreateSerializerSettings);            
+            _settings = new Lazy<Newtonsoft.Json.JsonSerializerSettings>(CreateSerializerSettings);
         }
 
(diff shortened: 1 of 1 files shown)


JW-CH:
> Filter by shared status: true = only shared, false = not shared, undefined = all owned albums

What are you trying to archieve here? Having shared not defined should result in showing both shared and not shared albums ([API docs](https://api.immich.app/endpoints/albums/getAllAlbums)).

Could you elaborate more?

JW-CH:
We are calling the `/Search/Random` endpoint which does not accept a `shared`-parameter. If you really want this, creating an issue in immich would be the best thing to do I assume.

3rob3:
> We are calling the `/Search/Random` endpoint which does not accept a `shared`-parameter. If you really want this, creating an issue in immich would be the best thing to do I assume.

This is only for albums, has nothing to do with search/random. You will still need to specify the album with this approach, I don't see any way around that.

JW-CH:
> This is only for albums, has nothing to do with search/random

Yes, but what are you trying to archieve?

3rob3:
> > Filter by shared status: true = only shared, false = not shared, undefined = all owned albums
> 
> What are you trying to archieve here? Having shared not defined should result in showing both shared and not shared albums ([API docs](https://api.immich.app/endpoints/albums/getAllAlbums)).
> 
> Could you elaborate more?

If you don't specify shared=true, you will not get back any shared albums (as that API doc explains). Try it:
- Create a NewUser and share an album with them
- As NewUser create your own new album and add some images
- In ImmichFrame config specify both albums UUIDs and NewUser's API key
- Result is only the NewUser images are shown, no shared album images.

JW-CH:
I think there might be a design flaw in immich but not a problem in ImmichFrame:

I have a album that is shared to me from another account. This album will be shown if I add id in immichframe.

The way the /albums API works rn:
shared=true: albums that I share and that are shared by others are shown
shared=false: albums that I don't share
shared=undefined: albums that are owned by me

so the behaviour of shared=true feels off but should not be an issue in ImmichFrame. A album that was shared is still displayed correctly in ImmichFrame.

JW-CH:
This will change in a future version of immich.

Quick confirmation from discord:
image

Issue #601, opened 2026-02-28
Stops Showing Media


duckdude555:
### Version

1.0.31.0

### Deployment Method

Android App 

### Description

After a period of time, ImmichFrame stops showing any new images - it hangs on an image with no progress bar. Force-closing the app and re-opening it fixes the issue until it happens again. I am running ImmichFrame as a "screensaver" app in Fully Kiosk on an Android Tablet. 

Happy to provide more information. Thanks!

### Reproduction

1. Set FullyKiosk to utilize immichframe app as screensaver app
2. Turn on fullyKiosk screensaver and wait

### Expectations

Immichframe should keep cycling through all pictures

### Configuration

```xml
eneral:
#  AuthenticationSecret: null
#  DownloadImages: false
#  RenewImagesDuration: 30
#  Webcalendars:
#    - calendarurl
#  RefreshAlbumPeopleInterval: 12
#  PhotoDateFormat: MM/dd/yyyy
#  ImageLocationFormat: 'City,State,Country'
  WeatherApiKey: ‘[api_key]’
  UnitSystem: imperial
  WeatherLatLong: ‘[lat_long]’
#  Webhook: null
  Language: en
  Interval: 45 
  TransitionDuration: 2
  ShowClock: true
  ClockFormat: 'hh:mm'
  ClockDateFormat: 'eee, MMM d'
  ShowProgressBar: true
  ShowPhotoDate: true
  ShowImageDesc: true
  ShowPeopleDesc: true
  ShowAlbumName: false
  ShowImageLocation: true
  #ShowVideos: true
  PrimaryColor: '#f5deb3'
  SecondaryColor: '#000000'
  Style: none
  BaseFontSize: 17px
  ShowWeatherDescription: true
  WeatherIconUrl: 'https://openweathermap.org/img/wn/{IconId}.png'
  ImageZoom: true
  ImagePan: false
  ImageFill: false
  Layout: splitview
Accounts:
  - ImmichServerUrl: http://[immich_ip]:2283
   # Exactly one of ApiKey or ApiKeyFile must be set.
    ApiKey: “[API_key]”
    # ApiKeyFile: "/path/to/api.key"
    ShowVideos: true
    ImagesFromDate: null
#    ShowMemories: false
#    ShowFavorites: false
#    ShowArchived: false
#    ImagesFromDays: null
#    ImagesUntilDate: '2020-01-02'
#    Rating: null
    Albums:
[...]

PR #634, opened 2026-04-20
Fix freeze caused by progress bar firing onDone during asset transitions

3rob3:
The progress bar's onChange handler was unconditionally calling play()
whenever the duration changed, causing onDone to fire multiple times
in rapid succession during asset transitions. This created a race
condition that caused a freeze.

Fixed by only calling play() in onChange when the progress bar is
already in a playing state. Also replaced the boolean
isHandlingAssetTransition guard with a promise-based transitionLock
to safely serialize any concurrent handleDone calls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset transition reliability by ensuring all transition operations execute in the correct sequence without interruption from concurrent requests.
  * Enhanced cleanup of video elements when components are destroyed to prevent playback issues and memory leaks.
  * Fixed event handlers from invoking after component destruction to prevent unexpected behavior and errors.


3rob3:
Closes #601

PR #651, opened 2026-05-30
feat(pool): exhaustive shuffle — show each asset once before repeating (resolves #438)

### LINKED ISSUE #438: Add exhaustive shuffle behavior so all images in an album are shown before reshuffling
### Description

Currently, when ImmichFrame displays images from an album, the selection order is random, and some images may be repeated frequently before others are shown. It would be helpful to implement an "exhaustive shuffle" mode where all images in the album are cycled through once before the set is reshuffled. This would ensure every photo is displayed at least once before any repeats.

### Use Case

This feature would improve the slideshow experience, especially for large albums, by ensuring all images are seen in a fair rotation. For example, if I have an album with 500 photos, I want to cycle through each of them before the order resets. This avoids the frustration of seeing the same photos multiple times while others are never shown.

### Details

- Add an option (e.g., "Exhaustive Shuffle" or "Cycle Through All Before Reshuffle") to the settings
- When enabled, ImmichFrame would generate a randomized list of images and display each one sequentially.
- Only after every image has been displayed would the app reshuffle and create a new randomized order.

### Pre-Submission Checklist

frslvr:
## What
Adds an opt-in `ExhaustiveShuffle` account setting so the slideshow shows every asset once before any repeats, instead of drawing each image independently at random. Closes #438.

## Why
Images are currently selected at random *with replacement* (`AllAssetsPool` uses Immich's random search; `CachingApiAssetsPool` reshuffles and takes per call), so some photos repeat many times before others appear at all — the behaviour reported in #438 (and #555).

## How
- New `ShuffleBagAssetPool : AggregatingAssetPool` decorates the source pool: it loads the set, shuffles it (Fisher–Yates), and serves each asset once; when the bag empties it refills from a fresh draw, so newly-added assets are picked up on the next cycle without a restart.
  - Dedupes by id, so multi-source aggregates don't double-serve.
  - Bag capped at 1000 to match Immich's random-search page limit; libraries larger than that become a rolling no-repeat window rather than the entire library per cycle.
- `ExhaustiveShuffle` (bool, default `false`) added to `IAccountSettings`, `ServerAccountSettings`, and the V1 settings + adapter, and wired into `PooledImmichFrameLogic.BuildPool`. Opt-in, so there is no behaviour change unless it is enabled.

## Scope / notes
- Single-client semantics: the bag lives in the (singleton) pool instance and is shared by all clients of an account. For a single frame this is exactly right; with several independent clients they share one rotation. This keeps the change small and matches the existing shared-state model — happy to extend to per-client bags if preferred.
- Exhaustive for single-source pools (all-assets, a single album, a person…); for multi-source aggregates it dedupes and is best-effort per cycle.

## Tests
- 5 new unit tests: no-repeat within a cycle, full-set coverage each cycle, dedup, empty pool, count delegation.
- Existing pool tests still pass.

## Summary by CodeRabbit

* **New Features**
  * Added an "Exhaustive Shuffle" option that, when enabled, shows each asset once in a randomized order before any repeats.

* **Tests**
  * Added tests validating exhaustive-shuffle behavior: full-cycle coverage, reshuffling between cycles, deduplication, empty-source handling, and max-bag capping.

* **Documentation**
  * Updated configuration docs and example settings to document the new Exhaustive Shuffle setting.


ViXXoR:
Looks like we're solving the same problem (https://github.com/immichFrame/ImmichFrame/pull/583).

This solution falls short where my first attempt did as well, in that all frames (if you have multiple) will show the same sort order.

Issue #614, opened 2026-03-28
Option to stream/display Original Images (Full Resolution)


fabb24:
### Description

Hi there,

First of all, thank you for this amazing project! It’s by far the best way to turn a screen into a dynamic photo frame with Immich.

I am currently using ImmichFrame on a large TV/high-resolution monitor. While the app works perfectly, I’ve noticed that the image quality is sometimes lacking. It seems that the app pulls the "Preview" or "Thumbnail" versions from Immich. 

### Use Case

On a 10-inch tablet, this is fine, but on a 65+ inch 4K screen, the compression artifacts and lack of sharpness are quite noticeable.

### Details

I would love to see a toggle or a parameter in the config.yaml (e.g., UseOriginalImages: true) that allows the app to fetch and display the original, uncompressed file instead of the generated thumbnails.

Even if this impacts loading times or bandwidth, having the option for users with local high-speed networks and large displays would drastically improve the visual experience (especially for high-megapixel photography).

Thank you for your all the hard work!

### Pre-Submission Checklist

- [x] This is a feature request and not a bug report

- [x] I have checked for related issues and checked the documentation

holisticode:
I would respectfully like to request the same feature. I am not expert in Immich, but if Immich tags the images when indexing, maybe the tag property can be used, but I am not sure it works. Thanks.

JW-CH:
There is already an open PR for this. I just linked it.

PR #652, opened 2026-06-02
feat: add full-resolution image option for asset loading

hacuka325:
Added new global setting UseFullResolutionImages with default false to preserve current behavior.
Updated image loading logic to request fullsize assets when the flag is enabled, otherwise keep using preview.

## Release Notes

* **New Features**
  * Added a configurable option to load full-resolution images instead of preview-sized images. The setting defaults to disabled and can be enabled through environment variables, JSON, or YAML configuration files.

* **Documentation**
  * Updated configuration documentation, Docker examples, and installation guides to include the new full-resolution images setting with default values and descriptions.


3rob3:
LGTM! Can you tell a difference visually? Tried on a lower end device to see if they struggle?
We should really drop Install_Web.md and Install_Client.md as those are super old and not used anymore. Not your problem but just noticed you edited one of them.

hacuka325:
Tried it on a Google TV Streamer with a 4K laser projector only, and it worked well. The difference isn't huge, but I believe I can see it :)

JW-CH:
Nice, clean change! Two small things:
  
Name: the other booleans here go without the Use prefix (ImageZoom, ImageFill, DownloadImages etc.), so FullResolutionImages or even FullsizeImages (matches the Immich API enum) would fit better.

Docs: worth adding a short note that full-res images can be way bigger than previews. Something like "note: this increases bandwidth and memory usage significantly".

Side note: if someone already has DownloadImages on and then enables this, they'll still get preview-sized images from the cache until they expire. Not a big deal, just good to know.

JW-CH:
Closes #614

PR #661, opened 2026-07-03
chore: gitignore '.@__thumb': qnap thumbnail cache

Zaida-3dO:
Adds two entries to `.gitignore`:

- `.@__thumb/` — QNAP NAS thumbnail cache directories (created by QNAP when the repo lives on a QNAP share)
- `.docker/` — Docker CLI / BuildKit local state (token seed, buildx node/refs) generated when `DOCKER_CONFIG` points at the project dir

## Summary by CodeRabbit

* **Chores**
  * Updated ignore rules to keep local settings and temporary cache files out of version control.
  * Added coverage for additional NAS thumbnail cache and Docker/BuildKit state directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -457,3 +457,9 @@ $RECYCLE.BIN/
 ## Ignore Settings file
 ImmichFrame/Settings.json
 ImmichFrame.WebApi/Config/Settings.json
+
+# QNAP NAS thumbnail cache
+.@__thumb/
+
+# Docker CLI / BuildKit local state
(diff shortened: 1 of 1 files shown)


JW-CH:
These entries are tool/OS-specific artifacts that are not generated by ImmichFrame itself. They belong in your personal global gitignore (`~/.gitignore_global`) rather than the project's `.gitignore`, which should only contain artifacts produced by the project itself (build outputs, generated files, etc.).

PR #662, opened 2026-07-03
Configurable asset fetch batch size

Zaida-3dO:
Assets are served from Immich via a stateless random search (`SearchRandomAsync`). Each batch returns 25 distinct assets, but there is no cross-batch "already shown" tracking — so once the frontend exhausts a batch and refetches, the next random draw can repeat images the viewer just saw. This means that currently, though I have hundreds of images in my album, I still get repetitions every 2 weeks or so.

### Change summary

Makes the number of assets fetched per request configurable via a new `AssetBatchSize` general setting (default 25, clamped to 1–1000), replacing the hardcoded 25 in `PooledImmichFrameLogic.GetAssets()`.

## Summary by CodeRabbit

* **New Features**
  * Added `General.AssetBatchSize` setting to control how many assets are fetched per request (default: 25), with matching updates to example Docker settings.
* **Bug Fixes**
  * Enforced safe batch-size limits by clamping the configured value to Immich’s supported range (1–1000).
* **Documentation**
  * Updated configuration reference and setup examples to include `General.AssetBatchSize`.
* **Tests**
  * Extended V1/V2 test configuration resources to cover the new setting.
* **Chores**
  * Updated version control ignore rules for additional generated/tool-local files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -457,3 +457,9 @@ $RECYCLE.BIN/
 ## Ignore Settings file
 ImmichFrame/Settings.json
 ImmichFrame.WebApi/Config/Settings.json
+
+# QNAP NAS thumbnail cache
+.@__thumb/
+
+# Docker CLI / BuildKit local state
+.docker/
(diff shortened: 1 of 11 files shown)


Zaida-3dO:
I just realized that this assumption is false if you use filters

> Each batch returns 25 distinct assets,

PR #694, opened 2026-08-15
Fix broken star history chart in README

FaintFlower:
The star history chart in the README is currently broken and renders nothing because the upstream service it relied on has shut down. This PR points it at a working replacement so the chart displays correctly again. No other changes are included.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Star History chart and link to use the new Star History website address.


3rob3:
Nice, thanks!

PR #706, opened 2026-09-16
Fix: WhereExcludes is O(n*m) and stalls the asset pool for large excluded albums

sheltonial:
## Problem

`WhereExcludes` rescans the entire excluded collection for every item in the source:

```csharp
public static IEnumerable<T> WhereExcludes<T>(this IEnumerable<T> source, IEnumerable<T> excluded, Func<T, object> comparator)
    => source.Where(item1 => !excluded.Any(item2 => Equals(comparator(item2), comparator(item1))));
```

That is O(n×m), and allocation-heavy with it: the `Func<T, object>` comparator boxes a `Guid` on every single comparison, on both sides.

It has one call site, `AssetExtensionMethods.ApplyAccountFilters`:

```csharp
assets = assets.WhereExcludes(excludedAlbumAssets, t => t.Id);
```

and `CachingApiAssetsPool` caches the resulting lazy `IEnumerable` rather than a materialised list, so the full scan re-runs on every enumeration of the pool.

With a small `ExcludedAlbums` this is invisible. On my library it isn't:

- **n = 30,313** assets across four configured `People`
- **m = 3,618** assets in an excluded album — videos over 30s, collected there so `ShowVideos` could be turned on without a 20-minute clip parking the frame for 20 minutes

That works out to ~5.5×10⁷ comparisons per enumeration, several enumerations per request. Authenticated `/api/Asset` requests stopped returning entirely — they hung past a 60s timeout, so the frame served nothing rather than erroring, which made it look like an empty pool rather than a slow one. Dropping back to the 98-asset excluded album I had before (~1.5×10⁶ comparisons) made it instant again.

## Fix

Hash the excluded keys once, then test membership:

```csharp
var excludedKeys = excluded.Select(comparator).ToHashSet();

return source.Where(item => !excludedKeys.Contains(comparator(item)));
```

O(n+m), and boxing drops from one per comparison to one per item. `TagAssetsPool` already uses a `HashSet<Guid>` for the same kind of lookup.

Semantics are unchanged: `EqualityComparer<object>.Default` calls the same `object.Equals` the old code did, and for a boxed `Guid` that is value equality backed by a consistent `GetHashCode`. Empty `excluded` and null elements behave as before.

One deliberate behavioural detail: since the body is no longer a single lazy expression, `excluded` is enumerated once when `WhereExcludes` is called, instead of repeatedly on each enumeration of the result. The only call site passes an already-materialised list, and building the set eagerly is precisely what makes repeat enumerations of the cached pool cheap.

Happy to go further and make the comparator generic (`Func<T, TKey>`) to remove boxing altogether, but that changes the public signature so I left it out of this PR.

## Tests

Adds `ImmichFrame.Core.Tests/Helpers/CollectionExtensionMethodsTests.cs`:

- excludes by comparator value, not instance identity
- empty `excluded` returns everything
- the result is stable when enumerated twice
- `excluded` is enumerated **once**, not once per source item

The last one is the regression guard. On `main` it fails with `Expected: 1, But was: 100`; the other three pass either way, which is the intended demonstration that the change is behaviour-preserving.

Existing `AllAssetsPoolTests.GetAssets_ExcludesAssetsFromExcludedAlbums` and the `AlbumAssetsPoolTests` exclusion cases cover the end-to-end semantics and are untouched.

Both suites pass locally on .NET 8: `ImmichFrame.Core.Tests` 88/88 and `ImmichFrame.WebApi.Tests` 26/26.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved collection filtering to compare items by their selected keys, producing more reliable exclusion results.
  - Improved efficiency when filtering large collections, reducing unnecessary repeated checks.
  - Ensured repeated enumeration returns consistent results and excluded data is processed only once.

- **Tests**
  - Added coverage for key-based exclusions, empty exclusion sets, repeated enumeration, and efficient collection processing.


3rob3:
Looks good to me, what do you think @JW-CH ?

Issue #669, opened 2026-07-09
ExcludedAlbums causes /api/asset to deadlock silently in v1.0.34 (cache poisoning wedges all clients); v1.0.33 crashes at the same code path


afonsolaw:
### Version

v1.0.34.0

### Deployment Method

Docker

### Description

1. Immich v3 server (v3.0.1) with a large library (~96,700 assets); create an album with a few thousand assets (~2,750 in my case).
2. Run ImmichFrame v1.0.34.0 (Docker) with ExcludedAlbums: "<that album uuid>". Also set in my config: People (3 person UUIDs), Interval, ShowClock - but removing ExcludedAlbums alone resolves it, and People alone works fine.
3. From a fresh client: curl -m 120 http://<frame>:8080/api/asset -> no response ever; curl exits with HTTP 000 at the timeout.
4. Container log shows the startup pool build completing normally (paginated POST /api/search/metadata, all 200s), then nothing: the hung /api/asset request produces ZERO outbound calls and ZERO errors.
5. Open the web UI from 1-2 more devices -> previously-working clients stop advancing too (their next asset request queues behind the poisoned cache entry). Web UI shows the "Looks like your immich-server is offline or you misconfigured immichFrame" page even though the server is healthy.
6. Remove ExcludedAlbums, recreate the container -> /api/asset returns within seconds and all clients work.

Reproduced identically on linux/amd64 (Synology DSM) and linux/arm64 (macOS/Colima), one via direct container network, one via HTTPS reverse proxy - not network- or architecture-specific.

### Reproduction

Immich v3 server with a large library; create an album with a few thousand assets.
Run ImmichFrame v1.0.34.0 with ExcludedAlbums: "<that album uuid>" (a People filter appears not to matter — see note above).
curl -m 120 http://<frame>:8080/api/asset from a fresh client → no response, curl exits with HTTP 000 at the timeout.
Container log shows the startup pool build completing normally (paginated POST /api/search/metadata, all 200s), then nothing: the hung /api/asset request produces zero outbound calls and zero errors.
Open the web UI from 1-2 more devices → previously-working clients stop advancing too (their next asset request queues behind the poisoned cache entry). Web UI shows the "Looks like your immich-server is offline or you misconfigured immichFrame" page even though the server is healthy.
Remove ExcludedAlbums, recreate the container → /api/asset returns within seconds and all clients work.

### Expectations

[...]

PR #707, opened 2026-09-17
Add HideAssetsInOtherAlbums account setting

adman234:
Adds a checkbox option to skip assets that are also in any album not listed under Albums, including albums shared with the user.

This is needed because sometimes photos can be put into multiple albums by accident when immich syncs them, and I don't want photos in my "documents" folder to be shown on my immichframe.

The Immich mobile app's album sync only ever adds assets to albums, so a photo moved between device albums stays in both server albums. Today the only fix is listing every other album in ExcludedAlbums and keeping that list current as new albums appear.

The other albums are resolved when the excluded-album cache is filled, so new albums are picked up on the next refresh. The setting has no effect without Albums, where every album would count as "other".

## Summary by CodeRabbit

- **New Features**
  - Added an account setting to hide assets that also appear in albums outside the selected album list, including shared albums.
  - Added an administration checkbox for configuring this behavior.
  - The setting applies only when specific albums are selected and defaults to disabled.

- **Documentation**
  - Added configuration guidance describing the setting and automatic updates when albums refresh.


adman234:
I should also mention:

**Related:** This works alongside #706 (the `WhereExcludes` speed fix). They touch different code and merge cleanly. Without #706, turning this on is as slow as a long `ExcludedAlbums` list (#669).

**Testing:** 4 new unit tests. On Linux CI in my fork, Core tests passed 88/88 and WebApi tests passed 26/26. In a benchmark with 8,000 selected assets and 30 other albums, one request took 81 s with this PR alone and 18 to 39 ms with #706 as well.

